PR #2440, opened 2023-12-15
feat: Move strings to resources for localization

LisoUseInAIKyrios:
Lays the groundwork for localizing patch strings.

See https://github.com/ReVanced/revanced-patches/pull/2593

[Integration changes](https://github.com/ReVanced/revanced-integrations/pull/420)


oSumAtrIX:
PR #2568 will likely conflict with this one, so it's fair to look into it and merge this PR before that. I'll look into the current merge conflicts.

src/main/kotlin/app/revanced/patches/music/misc/gms/YTMusicGmsCoreSupportResourcePatch.kt
```diff
@@ -1,11 +1,18 @@
 package app.revanced.patches.music.misc.gms
 
+import app.revanced.patcher.data.ResourceContext
 import app.revanced.patches.music.misc.gms.Constants.MUSIC_PACKAGE_NAME
 import app.revanced.patches.music.misc.gms.Constants.REVANCED_MUSIC_PACKAGE_NAME
+import app.revanced.patches.youtube.misc.strings.StringsPatch
 import app.revanced.patches.shared.misc.gms.AbstractGmsCoreSupportResourcePatch
 
-object GmsCoreSupportResourcePatch : AbstractGmsCoreSupportResourcePatch(
+object YTMusicGmsCoreSupportResourcePatch : AbstractGmsCoreSupportResourcePatch(
```
oSumAtrIX, line 9: If we want to include the app name in the class, I'd prefer to do this consistently for all eligible classes, ideally in another dedicated PR.

src/main/kotlin/app/revanced/patches/music/misc/strings/StringsPatch.kt
```diff
@@ -0,0 +1,7 @@
+package app.revanced.patches.music.misc.strings
+
+import app.revanced.patches.shared.strings.AbstractStringsResourcePatch
+
+object StringsPatch : AbstractStringsResourcePatch(
```
oSumAtrIX, line 5: Currently this patch is only a dependency of the Settings patch. This will be problematic in case we want to merge strings for a patch that does not depend on the Settings patch. Currently, this PR works for the reason that all patches depend on the Settings patch and therefor on the Strings patch. If this implicit dependency wouldn't exist, the StringsPatch would never execute, never close and thus never merge any strings.

src/main/kotlin/app/revanced/patches/twitch/ad/audio/AudioAdsPatch.kt
```diff
@@ -38,21 +38,13 @@ object AudioAdsPatch : BytecodePatch(
...
-                ),
-                StringResource(
-                    "revanced_block_audio_ads_on",
-                    "Audio ads are blocked"
-                ),
-                StringResource(
-                    "revanced_block_audio_ads_off",
-                    "Audio ads are unblocked"
-                ),
+                "revanced_block_audio_ads",
```
oSumAtrIX, line 45: Since we conventionally add string suffix to preference string keys, it may be a good idea to reduce boilerplate by just asking for a key for said preferences, and then simply infer the other strings it needs such as the title, summaries and so on. This way, each preference wouldn't need to reference so many strings anymore.

src/main/kotlin/app/revanced/patches/twitch/misc/translations/TranslationResourcePatch.kt
```diff
@@ -0,0 +1,26 @@
+package app.revanced.patches.twitch.misc.translations
+
+import app.revanced.patcher.data.ResourceContext
+import app.revanced.patcher.patch.ResourcePatch
+import app.revanced.patcher.patch.annotation.CompatiblePackage
+import app.revanced.patcher.patch.annotation.Patch
+import app.revanced.patches.twitch.misc.strings.StringsPatch
+
+@Patch(
+    //name = "Translations", // TODO: Once the patch is ready, uncomment this line.
```
oSumAtrIX, line 10: To me it feels odd that this patch is something the user can manually select. It feels much rather more than something that would be present automatically.

oSumAtrIX:
The PR introduces some changes regarding class names in integrations. I haven't touched them, but they should ideally not include the app name in the title following [this](https://github.com/ReVanced/revanced-integrations/pull/420#discussion_r1438771872).

oSumAtrIX:
PR https://github.com/ReVanced/revanced-patches/pull/2568 introduces a patch class to add strings and other app resources. Technically, this PR could use that PR by using the introduced patch class to add strings. Still, some distinct differences make it quite hard to make this PR work together with the other PR, which is quite unfortunate as this means we will have to deal with merge conflicts one way or another eventually.

The way the other PR works is as follows. When preferences are added via the settings patch, the resource callback for each preference when the preferences are added to the app. This callback uses the AddResourcesPatch, which adds the strings each preference uses. In a similar fashion, this PR could implement merging strings in the resource callback. The callback would return the keys for the strings used by the preferences and must be added to the app. In the callback, the strings patch from this PR would then look up all the strings that were returned in the callback on close and add them to the app.

This approach would eliminate the need to call StringsPatch.includeStrings() in every patch as the Settings patch would implicitly merge strings when it adds the preferences to the app, which in turn invoke the callback with the strings that are necessary for each preference. I'd say that this approach would also excuse adding all strings in a single XML file instead of separate XML files, just like it is done for the translation patch resources.

src/main/kotlin/app/revanced/patches/twitch/misc/translations/TranslationResourcePatch.kt
oSumAtrIX: Probably an option in the settings to use the native language instead of the translation would eliminate this problem

src/main/kotlin/app/revanced/patches/music/misc/strings/StringsPatch.kt
oSumAtrIX: Currently the only patch depending on StringsPatch is SettingsPatch which means every patch that currently adds strings, is also depending on SettingsPatch. So each patch that uses the StringsPatch, is missing a direct dependency to StringsPatch. In the case of adding strings  to say another app, they'd need to depend on the patch explicitly as they do not depend on a setting patch.

oSumAtrIX:
I have attempted to solve the merge conflicts, but I am sure I have missed some things. I consider rewriting the PR from scratch on top of the dev branch once [the other PR ](https://github.com/ReVanced/revanced-patches/pull/2568) is merged. This is because it's not worth merging only either, as they both contain changes that would be good to have present. I am not sure how to go about this but I'll try my best.

PR #2506, opened 2023-12-17
feat(Tiktok - Playback speed): Remember playback speed

d4rkk3y:
integration changes: https://github.com/ReVanced/revanced-integrations/pull/543
close: https://github.com/ReVanced/revanced-patches/issues/433


src/main/kotlin/app/revanced/patches/tiktok/interaction/speed/fingerprints/ChangeSpeedFingerprint.kt
```diff
@@ -0,0 +1,30 @@
+package app.revanced.patches.tiktok.interaction.speed.fingerprints
+
+import app.revanced.patcher.extensions.or
+import app.revanced.patcher.fingerprint.MethodFingerprint
+import com.android.tools.smali.dexlib2.AccessFlags
+import com.android.tools.smali.dexlib2.Opcode
+
+internal object ChangeSpeedFingerprint : MethodFingerprint(
```
oSumAtrIX, line 8: Are all attributes necessary for the fingerprint? If there is no particular reason, you may remove some attributes.

src/main/kotlin/app/revanced/patches/tiktok/interaction/speed/PlaybackSpeedPatch.kt
```diff
@@ -1,49 +1,89 @@
...
+    description = "Enables the playback speed option for all videos and " +
+            "retains the speed configurations in between videos.",
     compatiblePackages = [
         CompatiblePackage("com.ss.android.ugc.trill", ["32.5.3"]),
         CompatiblePackage("com.zhiliaoapp.musically", ["32.5.3"])
     ]
 )
 @Suppress("unused")
-object PlaybackSpeedPatch : BytecodePatch(setOf(SpeedControlParentFingerprint)) {
+object PlaybackSpeedPatch : BytecodePatch(
```
oSumAtrIX, line 31: This patch currently does two things at a same time. You may want to separate it, but its not necessary

src/main/kotlin/app/revanced/patches/tiktok/interaction/speed/PlaybackSpeedPatch.kt
```diff
@@ -1,49 +1,89 @@
...
         } ?: throw SpeedControlParentFingerprint.exception
+
+        GetSpeedFingerprint.result?.mutableMethod?.apply {
+            val injectIndex = indexOfFirstInstruction { getReference<MethodReference>()?.returnType == "F" } + 2
+            val register = getInstruction<Instruction11x>(injectIndex - 1).registerA
+
+            addInstruction(
+                injectIndex,
+                "invoke-static { v$register }," +
+                        " Lapp/revanced/tiktok/speed/SpeedPatch;->saveDefaultSpeed(F)V"
```
oSumAtrIX, line 69: You may call this method "rememberPlaybackSpeed"

src/main/kotlin/app/revanced/patches/tiktok/interaction/speed/PlaybackSpeedPatch.kt
```diff
@@ -1,49 +1,89 @@
...
-object PlaybackSpeedPatch : BytecodePatch(setOf(SpeedControlParentFingerprint)) {
+object PlaybackSpeedPatch : BytecodePatch(
+    setOf(
+        SpeedControlParentFingerprint,
+        GetSpeedFingerprint,
+        OnRenderFirstFrameFingerprint,
+        ChangeSpeedFingerprint
+    )
+) {
     override fun execute(context: BytecodeContext) {
```
oSumAtrIX, line 39: This patch has 3 blocks of code. Please add // region comments to document what each block of code does/ what its purpose is

src/main/kotlin/app/revanced/patches/tiktok/interaction/speed/PlaybackSpeedPatch.kt
```diff
@@ -1,49 +1,89 @@
...
+                        " Lapp/revanced/tiktok/speed/SpeedPatch;->saveDefaultSpeed(F)V"
+            )
+        } ?: throw GetSpeedFingerprint.exception
+
+        val changeSpeedMethod = ChangeSpeedFingerprint.result?.mutableMethod
+            ?: throw ChangeSpeedFingerprint.exception
+        OnRenderFirstFrameFingerprint.result?.mutableMethod?.addInstructions(
+            0,
+            """
+                const/4 v0, 0x1
```
oSumAtrIX, line 78: It is unknown what the registers are or what these instructions do. Please document it with comments (# Comment)

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/tiktok/interaction/speed/PlaybackSpeedPatch.kt
d4rkk3y: I think it is OK because these things support each other.

src/main/kotlin/app/revanced/patches/tiktok/interaction/speed/PlaybackSpeedPatch.kt
```diff
@@ -1,49 +1,88 @@
...
+        // Changes current video playback speed at the first frame of video using the saved playback speed.
+        // Because the default behavior of the TikTok app is that playback speed will reset to 1.0
+        // when swiping to the next video.
+        OnRenderFirstFrameFingerprint.result?.mutableMethod?.addInstructions(
+            0,
+            """
+                # The changeSpeedMethod have 3 arguments.
+                # First argument is a String. It changed depend on where video playing such as home page, following page, search result page, ...
+                # We are able to get it use getEnterFrom method.
+                const/4 v0, 0x1
```
oSumAtrIX, line 74: What is this constant? Why does it have to be 1?

src/main/kotlin/app/revanced/patches/tiktok/interaction/speed/PlaybackSpeedPatch.kt
```diff
@@ -1,49 +1,88 @@
...
+                invoke-virtual {p0, v0}, Lcom/ss/android/ugc/aweme/feed/panel/BaseListFragmentPanel;->getEnterFrom(Z)Ljava/lang/String;
+                move-result-object v0
+                # Second argument is a Aweme. It is some kind of data for a TikTok video.
```
oSumAtrIX, line 77: What is Aweme and what data is it? Why is it necessary to call `changeSpeedMethod`?

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/tiktok/interaction/speed/PlaybackSpeedPatch.kt
d4rkk3y: It is actually a boolean value. And I did not find out what it was. I set it to "true" because most of the time I see it set to "true" in decompiled code.

src/main/kotlin/app/revanced/patches/tiktok/interaction/speed/PlaybackSpeedPatch.kt
d4rkk3y: Aweme is a object which store data for a TikTok video. Here is some method of its class: 
https://github.com/ReVanced/revanced-integrations/blob/main/dummy/src/main/java/com/ss/android/ugc/aweme/feed/model/Aweme.java

d4rkk3y: I have just found a simpler approach without using that. I am testing it.

d4rkk3y: > I have just found a simpler approach without using that. I am testing it.

Nvm, it is not work.

oSumAtrIX: Why is the object necessary to change the object?

d4rkk3y: > Why is the object necessary to change the speed?

Afaik, when the handle play speed changed event, app will call Aweme.getAid() from that object to get the video identifier and compare it with the getCurrentAweme().getAid(). If they are different,  app will set the speed to 1.0f.
image

oSumAtrIX:
I have pushed some changes. Please check if it looks good and if the names for the fingerprints I inferred from the code is correct.

d4rkk3y:
> I have pushed some changes. Please check if it looks good and if the names for the fingerprints I inferred from the code is correct.

I tested it, and it worked well. But I think the fingerprint's name should not be `OnVideoSwiped` because it literally resolve to the method used to change video speed.

oSumAtrIX:
Integrations aren't merged yet

PR #2512, opened 2023-12-18
feat(YouTube - Change header): Change to ReVanced header by default

oSumAtrIX:
Changes the default header to:

image


src/main/kotlin/app/revanced/patches/youtube/layout/branding/header/ChangeHeaderPatch.kt
```diff
@@ -0,0 +1,177 @@
...
+import app.revanced.patcher.patch.annotation.Patch
+import app.revanced.patcher.patch.options.PatchOption.PatchExtensions.stringPatchOption
+import app.revanced.util.ResourceGroup
+import app.revanced.util.copyResources
+import app.revanced.util.inputStreamFromBundledResource
+import java.awt.Graphics2D
+import java.awt.GraphicsConfiguration
+import java.awt.GraphicsEnvironment
+import java.awt.Transparency
+import java.awt.image.BufferedImage
```
oSumAtrIX, line 15: Unfortunately, I believe this API does not exist in the Android framework.

oSumAtrIX: The correct approach would be to use Kotlin Multiplatform, but before this got stable, https://github.com/ReVanced/revanced-patches/tree/main/dummy was used as a stub of the Android API to compile the patches. Alternatively, a `compileOnly` dependency can be added to the local Android `framework.jar` file or https://mvnrepository.com/artifact/com.android/framework/8.1 could be used.

oSumAtrIX: Regarding to implementation, an `AndroidImageProcessor : ImageProcessor` can be created that overrides the `ImageProcessor` JWT implementation with that of an Android one.

oSumAtrIX: An alternative way is to ask YouTube the intended way ™️ to use their brand assets. We have done this in the past with success as expected. In this case, it may be slightly overkill, and I'd much rather not include the option and provide what we have now, which is:

- YouTube Premium
- YouTube
- ReVanced
- Custom

This would spare shipping an entire image processing library just to support multiplatform and one additional header (ReVanced YouTube), which the user can technically use with the custom option, but if you think its a good addition, sure.

src/main/kotlin/app/revanced/patches/youtube/layout/branding/header/ChangeHeaderPatch.kt
```diff
@@ -47,12 +38,12 @@ object ChangeHeaderPatch : ResourcePatch() {
...
     private val header by stringPatchOption(
         key = "header",
-        default = "ReVanced YouTube",
+        default = "ReVanced minimal",
         values = mapOf(
             "YouTube" to HEADER_NAME,
             "YouTube Premium" to PREMIUM_HEADER_NAME,
             "ReVanced" to REVANCED_HEADER_NAME,
-            "ReVanced YouTube" to REVANCED_YOUTUBE_HEADER_NAME,
+            "ReVanced minimal" to REVANCED_MINIMAL_HEADER_NAME,
```
oSumAtrIX, line 46: Maybe call this "ReVanced (borderless icon)," so it is clear what the difference is to "ReVanced". I am conflicted about the header as we do not feature such a header without a border in our branding assets, which is why it'd be slightly inconsistent.

oSumAtrIX (commented):
I am conflicted about the header with the borderless icon as we do not feature such a header in our branding assets, which is why it'd be slightly inconsistent. Any input on this?

oSumAtrIX:
@Ushie What is your input on the defaults. We received a suggestion to not change the branding of YouTube by default which is why we disabled the custom branding patch back then. This patch is enabled by default and changes the header to ReVanced by default. Previously it changed it to premium by default but we got input to not do that as it may convey a false meaning that YouTube Premium is now active.

@LisoUseInAIKyrios you had some plans regarding merging the custom branding and custom header patch, if this is still something that makes sense, as the custom branding patch is disabled by default, so would this patch be

Ushie:
> What is your input on the defaults. We received a suggestion to not change the branding of YouTube by default which is why we disabled the custom branding patch back then. This patch is enabled by default and changes the header to ReVanced by default. Previously it changed it to premium by default but we got input to not do that as it may convey a false meaning that YouTube Premium is now active.

I don't believe any part of the app design should be changed by default, after all it is YouTube, it should use the YouTube logo and the YouTube official headers, the same reason Custom Branding by default doesn't change the app icon

As for the premium part, I agree it should be changed, not to change anything by default, ideally it could be an in-app setting but it's a resource patch so I don't know how viable that is

Ushie:
> Added another patch option to use the ReVanced logo with the original "YouTube" header text. The logo uses the borderless icon and a similar layout to the original YT icon.
> 
> revanced_youtube_light revanced_youtube_dark

Can't this be mistaken for an official ReVanced header? it reinforces the patched YouTube = ReVanced misconception

oSumAtrIX:
The same way the Custom branding patch uses the ReVanced Logo this patch does too

Ushie:
> The same way the Custom branding patch uses the ReVanced Logo this patch does too

Not in the same way, app name and app icon are placed separately, this is presented in the form of a wordmark

PR #2526, opened 2023-12-21
feat(ID Austria): Remove constraint on any version

1fexd:


oSumAtrIX (commented):
Maybe the version constraint can be removed?

1fexd:
Nope, the patch only works with `3.x.x`, not with `2.x.x`

oSumAtrIX:
The idea is to support the latest and future versions

1fexd:
Is there any way to do a semver check? That would be cool

oSumAtrIX:
The version field is being compared against the versionName which can contain any raw string: 
image

We would need to compare against `versionCode` but then it is not really user friendly.

PR #2540, opened 2023-12-25
feat(Twitter): Add `Change link sharing domain` patch

FrozenAlex:
The patch replaces x.com and twitter.com in share by fxtwitter.com and optionally other mirrors.
Since twitter does not have a settings GUI yet, I use patch options to configure the patch.
This pull request also contains `Remove tracking query parameter` patch

### DIFF
--- a/src/main/kotlin/app/revanced/patches/twitter/misc/links/ChangeLinkSharingDomainPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/twitter/misc/links/ChangeLinkSharingDomainPatch.kt
@@ -42,10 +42,10 @@ object ChangeLinkSharingDomainPatch : BytecodePatch(
     )
 
     override fun execute(context: BytecodeContext) {
-        // Replace the domain in the method that generates the share link
-        // This is used in the copy link button to generate the link that is copied to the clipboard
+        // Replace the domain in the method that generates the share link.
+        // This is used in the copy link button to generate the link that is copied to the clipboard.
         val linkBuilderMethod = LinkBuilderMethodFingerprint.result?.let {
-            // Find the index of the string instruction that contains the link
+            // Find the index of the string instruction that contains the link.
             val stringIndex = it.scanResult.stringsScanResult!!.matches.find { match ->
                 (match.string == "https://twitter.com/%1\$s/status/%2\$d" || match.string == "https://x.com/%1\$s/status/%2\$d")
             }!!.index
@@ -78,33 +78,33 @@ object ChangeLinkSharingDomainPatch : BytecodePatch(
         LinkResourceGetterFingerprint.result?.apply {
             val instructions = this.mutableMethod.getInstructions()
 
-            // Result register of the original method call
+            // Result register of the original method call.
             var resultRegister = 0
 
-            // Remove instructions at the end and get the register of the last instruction
+            // Remove instructions at the end and get the register of the last instruction.
             for ((index, instruction) in instructions.withIndex()) {
                 if (instruction.opcode == Opcode.INVOKE_VIRTUAL) {
                     val methodRef =
                         (instruction as ReferenceInstruction).reference as MethodReference
                     if (methodRef.definingClass != "Landroid/content/res/Resources;") continue
 
-                    // Get the result register of the original method call
+                    // Get the result register of the original method call.
                     resultRegister = (instructions[index + 1] as OneRegisterInstruction).registerA
 
-                    // Remove the original method call for getting the link from resources and the move-result-object instruction
+                    // Remove the original method call for getting the link from resources and the move-result-object instruction.
                     this.mutableMethod.removeInstructions(index, 2)
 
-                    // Remove the instruction that uses the resultRegister as an array reference to prevent an error
+                    // Remove the instruction that uses the resultRegister as an array reference to prevent an error.
                     this.mutableMethod.removeInstructions(index - 2, 1)
 
                     break
                 }
             }
 
-            // Instruction that sets free register to "this", needed to restore the original value of the register
+            // Instruction that sets free register to "this", needed to restore the original value of the register.
             var instructionToGetThis: TwoRegisterInstruction? = null
 
-            // Save user nickname to free register
+            // Save user nickname to free register.
             for ((index, instruction) in instructions.withIndex()) {
                 if (instruction.opcode == Opcode.INVOKE_VIRTUAL) {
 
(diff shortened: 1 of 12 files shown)


src/main/kotlin/app/revanced/patches/twitter/misc/links/ChangeLinkSharingDomainPatch.kt
```diff
@@ -0,0 +1,148 @@
...
+                convertTweetIdToLongIndex,
+                """
+                    invoke-static { v$tweetIdP1, v$tweetIdP2, v$tempNicknameRegister }, $buildShareLinkMethod
+                    move-result-object v$shareLinkRegister
+                """
+            )
+
+            // Restore the borrowed nicknameRegister.
+            addInstruction(
+                // TODO: Comment this offset.
```
oSumAtrIX, line 140: Please comment this offset

src/main/kotlin/app/revanced/patches/twitter/misc/links/RemoveTrackingQueryParameterPatch.kt
```diff
@@ -0,0 +1,24 @@
...
+    name = "Remove tracking query parameter",
+    description = "Remove the tracking query parameter from links.",
+    compatiblePackages = [CompatiblePackage("com.twitter.android")]
+)
+@Suppress("unused")
+object RemoveTrackingQueryParameterPatch : BytecodePatch(
+    setOf(AddTrackingQueryToLinkFingerprint)
+) {
+    override fun execute(context: BytecodeContext) {
+        AddTrackingQueryToLinkFingerprint.result?.mutableMethod?.addInstruction(0, "return-object p0")
```
oSumAtrIX, line 21: What is p0?

src/main/kotlin/app/revanced/patches/twitter/misc/links/ChangeLinkSharingDomainPatch.kt
```diff
@@ -0,0 +1,148 @@
...
+import com.android.tools.smali.dexlib2.iface.instruction.Instruction
+import com.android.tools.smali.dexlib2.iface.instruction.OneRegisterInstruction
+import com.android.tools.smali.dexlib2.iface.instruction.TwoRegisterInstruction
+import com.android.tools.smali.dexlib2.iface.reference.MethodReference
+
+
+@Patch(
+    name = "Change link sharing domain",
+    description = "Replaces the domain name of Twitter links when sharing them.",
+    compatiblePackages = [CompatiblePackage("com.twitter.android")]
```
oSumAtrIX, line 31: Please check if this patch still works with the latest version of Twitter. The patch uses quite a lot of offset constants and may stop working due to that quite easily.

src/main/kotlin/app/revanced/patches/twitter/misc/links/ChangeLinkSharingDomainPatch.kt
```diff
@@ -0,0 +1,148 @@
...
+            val tweetIdP2 = tweetIdP1 + 1
+
+            replaceInstructions(
+                // This offset is used to place the instruction after the save of the method result.
+                convertTweetIdToLongIndex,
+                """
+                    invoke-static { v$tweetIdP1, v$tweetIdP2, v$tempNicknameRegister }, $buildShareLinkMethod
+                    move-result-object v$shareLinkRegister
+                """
+            )
```
oSumAtrIX, line 136: Quite a lot happens here. Is it really necessary to patch this specific method? It is quite low level, maybe there's a better location to patch

FrozenAlex: The alternative to that would be changing the string in resources, but I was advised against it, cause we won't be able to change the string resources at runtime. Twitter has two different ways to construct a string for sharing. This one in particular is using a string in resources as a template, and is used when twitter shows the text of the sharing link to the user. In this section I just make it so the code runs the method I patched earlier, that has the string template in the code itself.
But it would be much more stable if we had it patch resources, since it does not require all of these modifications. And settings menu for twitter is not implemented anyway. If I would make a stable patch, I would just patch resources and remove this entire patching section

FrozenAlex: The other way to solve it would be to create an integration that has a method that receives all needed types, since half of the complexity of this code comes from converting all the types for our first patched method to be called

FrozenAlex: I think for now, making it patch a string in resources would be the best solution since if we make an integration, we don't need type conversions here anyway and it will greatly reduce the complexity of the patch

src/main/kotlin/app/revanced/patches/twitter/misc/links/ChangeLinkSharingDomainPatch.kt
```diff
@@ -0,0 +1,148 @@
...
+        required = true
+    )
+
+    override fun execute(context: BytecodeContext) {
+        fun Method.indexOfFirstOrThrow(errorMessage: String, predicate: Instruction.() -> Boolean) =
+            indexOfFirstInstruction(predicate).also {
+                if (it == -1) throw PatchException(errorMessage)
+            }
+
+        // region Replace the domain name when copying a link with "Copy link" button.
```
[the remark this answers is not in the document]
FrozenAlex: Yeah it is correct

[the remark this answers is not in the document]
FrozenAlex: Yeah it is correct

src/main/kotlin/app/revanced/patches/twitter/misc/links/RemoveTrackingQueryParameterPatch.kt
FrozenAlex: p0 is the sharing url, the original function receives shareLink, shareParam, sessionToken and adds tracking parameters based on the shareParam and sessionToken to the shareLink and returns it in the end. In this patch we just return the original string, which is passed as p0 at the start of the function

oSumAtrIX: please document this in the code

src/main/kotlin/app/revanced/patches/twitter/misc/links/ChangeLinkSharingDomainPatch.kt
oSumAtrIX: I think we came up with a clean solution in Discord, is this correct

oSumAtrIX:
Is this PR still active?

PR #2541, opened 2023-12-25
refactor: Fix package and code structure

oSumAtrIX:
RE: https://github.com/ReVanced/revanced-integrations/pull/546

# TODO

- [x] Keep an eye on the changes involved in PR #2568


src/main/kotlin/app/revanced/patches/shared/integrations/AbstractIntegrationsPatch.kt
```diff
@@ -53,11 +71,7 @@ abstract class AbstractIntegrationsPatch(
...
     }
 
-    override fun execute(context: BytecodeContext) {
-        if (context.findClass(integrationsDescriptor) == null) throw PatchException(
-            "Integrations have not been merged yet. This patch can not succeed without merging the integrations."
-        )
-
-        for (hook in hooks) hook.invoke(integrationsDescriptor)
+    private companion object {
+        private const val INTEGRATIONS_CLASS_DESCRIPTOR = "Lapp/revanced/integrations/shared/Utils;"
```
johnconner122, line 75:
```suggestion
        const val INTEGRATIONS_CLASS_DESCRIPTOR = "Lapp/revanced/integrations/shared/Utils;"
```

oSumAtrIX: Why was this made public?

johnconner122: There is already private modifier with `companion object`.

oSumAtrIX:
PR #2568 makes conflicting changes with this PR. Some strings involved in the other PR will have to be updated to match the changes in this PR if this PR is merged at first.

src/main/kotlin/app/revanced/patches/shared/integrations/AbstractIntegrationsPatch.kt
oSumAtrIX: There is a difference. The visibility of the companion object does not apply to the field. If you change the field modifier you will actually change the visibility for the field and add it to the public API:

image

PR #2551, opened 2023-12-28
feat(YouTube): Support version `18.49.36`

oSumAtrIX:
# Todo

- [ ] Fix RYD not resolving (once again, we really need better fingerprints)
- [ ] Fix Playback speeds not resolving
- [ ] Check which patches lose compatibility with older versions and drop support
- [ ] Bump compatibility


oSumAtrIX:
I haven't looked into it yet, but in case it blocks the bump we can do so without RYD at first

oSumAtrIX:
In this case, I can understand how this could be cumbersome, though, generally speaking, there is no relationship between these patches, so they can be bumped independently without any consideration toward each other, as all the patches specify their compatibility on their own which is where I came from with my last comment. We can wait until we can figure out a solution for RYD or retain compatibility with existing versions. In the future, if we want to create such a relationship between patches, we need to explicitly provide a design or system for that.

Issue #2576, opened 2024-01-01
bug (YouTube - Settings): Setting titles get cut off when they can't fit on screen


KobeW50:
### Bug description

Setting titles don't get line-wrapped and therefore get cut off if they don't fit the screen.

![SmartSelect_20240101_135606](https://github.com/ReVanced/revanced-patches/assets/84587632/964a0d7d-d4c0-48c4-8535-a0f6dc0dd814)

![Screenshot_20240101_135149](https://github.com/ReVanced/revanced-patches/assets/84587632/be49399b-6979-4ad7-87e2-158854373ca8)

### Solution

Two possible solutions:

A) Wrap titles onto new lines just like descriptions.

B) Use shorter titles by convention.

### Acknowledgements

- [X] This issue is not a duplicate of an existing bug report.
- [X] I have chosen an appropriate title.
- [X] All requested information has been provided properly.

KobeW50:
> Maybe rename
> `Hide search result recommendations (e.g People also watched)`
> to something like
> `Hide 'People also watched' recommendations`

Alternatively, it can be `Hide search result recommendations`, and then the example can be given in the description.

> `show a toast when skipping automatically` can still be understood even with the last bit clipped, as it's fully described in the summary.  But maybe there's a suitable shorter title that also works.

Maybe `Notify when skipping automatically`? The description will describe that it is a toast that notifies.

PR #2579, opened 2024-01-02
fix(YouTube): Shorten setting titles to fit on screen

KobeW50:
Resolves https://github.com/ReVanced/revanced-patches/issues/2576

### DIFF
--- a/src/main/kotlin/app/revanced/patches/youtube/layout/hide/general/HideLayoutComponentsPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/youtube/layout/hide/general/HideLayoutComponentsPatch.kt
@@ -97,15 +97,15 @@ object HideLayoutComponentsPatch : BytecodePatch(
                 "revanced_hide_search_result_recommendations",
                 StringResource(
                     "revanced_hide_search_result_recommendations_title",
-                    "Hide search result recommendations (e.g People also watched)"
+                    "Hide search result recommendations"
                 ),
                 StringResource(
                     "revanced_hide_search_result_recommendations_summary_on",
-                    "Recommendations are hidden"
+                    "Recommendations such as People also watched are hidden"
                 ),
                 StringResource(
                     "revanced_hide_search_result_recommendations_summary_off",
-                    "Recommendations are shown"
+                    "Recommendations such as People also watched are shown"
                 )
             ),
             SwitchPreference(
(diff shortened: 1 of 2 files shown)


src/main/resources/sponsorblock/host/values/strings.xml
```diff
@@ -13,7 +13,7 @@
     <string name="sb_enable_auto_hide_skip_segment_button">Automatically hide skip button</string>
     <string name="sb_enable_auto_hide_skip_segment_button_sum_on">Skip button hides after a few seconds</string>
     <string name="sb_enable_auto_hide_skip_segment_button_sum_off">Skip button displayed for entire segment</string>
-    <string name="sb_general_skiptoast">Show a toast when skipping automatically</string>
+    <string name="sb_general_skiptoast">Notify when skipping automatically</string>
```
oSumAtrIX, line 16: Notify can mean anything. Toast is better here.

KobeW50: So just `Toast when skipping automatically`?

Can the verb "toast" be used this way?

src/main/kotlin/app/revanced/patches/youtube/layout/hide/general/HideLayoutComponentsPatch.kt
```diff
@@ -97,15 +97,15 @@ object HideLayoutComponentsPatch : BytecodePatch(
                 "revanced_hide_search_result_recommendations",
                 StringResource(
                     "revanced_hide_search_result_recommendations_title",
-                    "Hide search result recommendations (e.g People also watched)"
+                    "Hide search result recommendations"
                 ),
                 StringResource(
                     "revanced_hide_search_result_recommendations_summary_on",
-                    "Recommendations are hidden"
+                    "Recommendations such as People also watched are hidden"
```
oSumAtrIX, line 104: Moving that to the description contributes to inconsistency in description style. Not sure how I feel about this. If we would want to include this here, we would want to do this consistently for all descriptions.

KobeW50: I also find it a bit offputting.

I don't think this will need to be added to all descriptions. Only descriptions that need an example for additional context.

Other descriptions already do similar things, such as the Hide YouTube Premium promotions description.

oSumAtrIX: As said in my previous comment, it contributes to the cause. We have some consistency problems.

src/main/resources/sponsorblock/host/values/strings.xml
oSumAtrIX: The original string would be used here. Notify is a synonym for show a toast. It is better to use a known set of terms instead of introducing variations.

oSumAtrIX (approved):

Issue #2580, opened 2024-01-02
bug(YouTube - Change header): Patch should be named "Custom header" and string improvements needed 


KobeW50:
### Bug description

1. The name `Change header` can be improved. 

2. The patch description should follow the wording pattern of similar patches (such as the Custom branding patch)

3. The first two sentences of the patch options description are difficult to understand.

### Solution

1. The patch should be renamed to `Custom header`. This is more appropriate and better describes what it does.

2. The patch description can be changed to `Applies a custom header in the top left corner within the app. Defaults to the ReVanced header.`

3. The option description can be changed to `Can be a header name or a path to a custom header folder to use in the top bar. A path to a folder must contain one or more of the following folders matching the DPI of your device: ...`
            
            

### Additional context

I can PR for this. Here is how the changes for 2 and 3 look:

![image](https://github.com/ReVanced/revanced-patches/assets/84587632/bdbe7f36-b228-4ea0-88f2-c534befef413)

### Acknowledgements

- [X] This issue is not a duplicate of an existing bug report.
- [X] I have chosen an appropriate title.
- [X] All requested information has been provided properly.

PR #2581, opened 2024-01-03
fix(YouTube - Change header): Improve patch descriptions

KobeW50:
Closes https://github.com/ReVanced/revanced-patches/issues/2580

### DIFF
--- a/src/main/kotlin/app/revanced/patches/youtube/layout/branding/header/CustomHeaderPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/youtube/layout/branding/header/CustomHeaderPatch.kt
@@ -11,7 +11,7 @@ import app.revanced.util.copyResources
 import java.io.File
 
 @Patch(
-    name = "Change header",
+    name = "Custom header",
     description = "Change the in app header. Defaults to the ReVanced header.",
     compatiblePackages = [
         CompatiblePackage("com.google.android.youtube")
@@ -19,7 +19,7 @@ import java.io.File
     use = false
 )
 @Suppress("unused")
-object ChangeHeaderPatch : ResourcePatch() {
+object CustomHeaderPatch : ResourcePatch() {
     private const val HEADER_NAME = "yt_wordmark_header"
     private const val PREMIUM_HEADER_NAME = "yt_premium_wordmark_header"
     private const val REVANCED_HEADER_NAME = "ReVanced"
--- a/src/main/kotlin/app/revanced/patches/youtube/layout/branding/header/CustomHeaderPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/youtube/layout/branding/header/CustomHeaderPatch.kt
@@ -12,7 +12,7 @@ import java.io.File
 
 @Patch(
     name = "Custom header",
-    description = "Change the in app header. Defaults to the ReVanced header.",
+    description = "Applies a custom header in the top left corner within the app. Defaults to the ReVanced header.",
     compatiblePackages = [
         CompatiblePackage("com.google.android.youtube")
     ],
@@ -48,8 +48,8 @@ object CustomHeaderPatch : ResourcePatch() {
         ),
         title = "Header",
         description = """
-            The header to use in top bar or the path to a custom header.
-            The path to a folder containing one or more of the following folders matching the DPI of your device:
+            Can be a header name or a path to a custom header folder to use in the top bar.
+            A path to a folder must contain one or more of the following folders matching the DPI of your device:
 
             ${targetResourceDirectoryNames.joinToString("\n") { "- $it" }}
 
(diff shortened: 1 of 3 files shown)


src/main/kotlin/app/revanced/patches/youtube/layout/branding/header/CustomHeaderPatch.kt
```diff
@@ -48,8 +48,8 @@ object ChangeHeaderPatch : ResourcePatch() {
         ),
         title = "Header",
         description = """
-            The header to use in top bar or the path to a custom header.
-            The path to a folder containing one or more of the following folders matching the DPI of your device:
+            Can be a header name or a path to a custom header folder to use in the top bar.
```
oSumAtrIX, line 51: "Header name" is kinda difficult to understand

```suggestion
            Either a header name or a path to a custom header folder to use in the top bar.
```

src/main/kotlin/app/revanced/patches/youtube/layout/branding/header/CustomHeaderPatch.kt
```diff
@@ -48,8 +48,8 @@ object ChangeHeaderPatch : ResourcePatch() {
...
+            A path to a folder must contain one or more of the following folders matching the DPI of your device:
```
oSumAtrIX, line 52: We refer to the specific path if used for the option.
```suggestion
            The path to a folder must contain one or more of the following folders matching the DPI of your device:
```

src/main/kotlin/app/revanced/patches/youtube/layout/branding/header/CustomHeaderPatch.kt
```diff
@@ -11,15 +11,15 @@ import app.revanced.util.copyResources
 import java.io.File
 
 @Patch(
-    name = "Change header",
-    description = "Change the in app header. Defaults to the ReVanced header.",
+    name = "Custom header",
```
oSumAtrIX, line 14: Patches named by their action is better. "Change header" says what this patch does. Revert the changes regarding the patch name

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/youtube/layout/branding/header/CustomHeaderPatch.kt
KobeW50: Many patches don't follow this rule. This patch is very similar to `Custom branding` which is a reason to give them consistent names. Also, "custom" tells the user what they can do with the patch while "change" is ambiguous.

KobeW50:
If the patch name would change, in lines 95 and 102 of CustomHeaderPatch.kt there is a bit of code that is still named change-header

I didn't change this to custom-header because I wasn't sure what it does.

https://github.com/ReVanced/revanced-patches/commit/b596b2c37b8a2766bc35dbf5d30675fa07f68656#diff-5e83a02d14cb9cd781c7acbde6850469bc0fb34b637cb6a8fd0d1fa798819b6dL95-R102

src/main/kotlin/app/revanced/patches/youtube/layout/branding/header/CustomHeaderPatch.kt
oSumAtrIX: Like I said, it contributes to the problem, just because other patches contribute to it, doesn't excuse this patch to too.

oSumAtrIX:
Keep the patch name as is, don't change it to custom header. You'd also need to commit the API changes besides

src/main/kotlin/app/revanced/patches/youtube/layout/branding/header/CustomHeaderPatch.kt
KobeW50: > "Header name" is kinda difficult to understand

Perhaps all patch option presets should be referred to as a "preset" or synonymous term

Pamilg8:
Screenshot_20240103-135842
A patch to change that too. Pls

src/main/kotlin/app/revanced/patches/youtube/layout/branding/header/CustomHeaderPatch.kt
oSumAtrIX: This terminology would be introduced here and nowhere else which would be inconsistent. If you want this to change, do so please in a separate PR consistently

KobeW50: It's fine for now. Maybe in the future if it's needed.

I don't have a better term than "header name" so I believe this PR is ready to be merged.

oSumAtrIX (approved):

PR #2642, opened 2024-01-26
Configure Dependabot

CanerKaraca23:


.github/dependabot.yml
```diff
@@ -6,17 +6,53 @@ updates:
     target-branch: dev
     schedule:
       interval: monthly
+    commit-message:
+      prefix: fix
```
oSumAtrIX, line 10: Why is the prefix "fix"? Shouldn't it be "build"?

.github/dependabot.yml
```diff
@@ -6,17 +6,53 @@ updates:
...
+      prefix-development: build
+      include: scope
```
oSumAtrIX, line 12: We don't use any scopes unless we need the dependency bumps to make a release. For that we use the "Needs bump" scope. WHen we need it we can manually squash the PR to include the scope, but by default no scope should be included

.github/dependabot.yml
```diff
@@ -6,17 +6,53 @@ updates:
...
+    groups:
+      actions-dependencies:
+        update-types:
+        - "minor"
```
oSumAtrIX, line 16: What does this configuration do?

oSumAtrIX (changes requested):

.github/dependabot.yml
CanerKaraca23: > Why is the prefix "fix"? Shouldn't it be "build"?

Dependabot detects if it's production dependency or dev dependency. If it's production, it's using "fix", if it's dev, it's using "build". We can change it to "build" and remove "prefix-development" if you want. @oSumAtrIX

.github/dependabot.yml
CanerKaraca23: > What does this configuration do?

Will mark dependencies if it's major, minor or patch update.

.github/dependabot.yml
oSumAtrIX: Why would it be a fix? What does it fix? It only bumps a dependency. The prefix can be removed, yes

.github/dependabot.yml
oSumAtrIX: What do you mean mark?

CanerKaraca23: > What do you mean mark?

It will only say if it's major, minor or patch on description.

oSumAtrIX: What is major, minor or patch? The dependency version? The version after the bump of ReVanced Patches?

CanerKaraca23: > What is major, minor or patch? The dependency version? The version after the bump of ReVanced Patches?

Yes, dependency version. For example, "gson" = 2.10.1 to 3.0.0 is a major update. It only tells this to you. Major updates may break build. It's just a simple thing.

.releaserc
```diff
@@ -9,12 +9,31 @@
...
+            },
+            {
+              "type": "fix",
+              "section": "Bug Fixes"
+            },
+            {
+              "type": "build",
+              "section": "Dependencies and Other Build Updates",
+              "hidden": false
+            }
```
oSumAtrIX, line 35: Is it possible to specify only the build type? Features and fixes are already being added to the changelog.

CanerKaraca23: > Is it possible to specify only the build type? Features and fixes are already being added to the changelog.

That "hidden": false, if you don’t include it, it’ll default to true, so already hidden.

oSumAtrIX: This is not what I am asking. Whats the impact of removing line 23-30

.github/dependabot.yml
oSumAtrIX: > "gson" = 2.10.1 to 3.0.0 is a major update. It only tells this to you. 

Then why are line 14 and 15 included?

oSumAtrIX (changes requested):

PR #2697, opened 2024-02-07
feat(OpeningHours): Add `Fix crash` patch

Linus789:
Old PR: https://github.com/ReVanced/revanced-patches-template/pull/3271

As LisoUseInAIKyrios suggested I contacted the developer of the app. He said that it was a hobby project and there won't be any updates. I asked if he could make the source code public, but I haven't received a reply and several months have already passed.

And thanks oSumAtrIX for letting me know of the new repository. As you suggested I have re-opened the PR here.


src/main/kotlin/app/revanced/patches/openinghours/crash/fingerprints/SetPlaceFingerprint.kt
```diff
@@ -0,0 +1,12 @@
+package app.revanced.patches.openinghours.crash.fingerprints
+
+import app.revanced.patcher.fingerprint.MethodFingerprint
+
+object SetPlaceFingerprint : MethodFingerprint(
```
oSumAtrIX, line 5:
```suggestion
internal object SetPlaceFingerprint : MethodFingerprint(
```

src/main/kotlin/app/revanced/patches/openinghours/crash/CrashPatch.kt
```diff
@@ -0,0 +1,86 @@
...
+                        move-result-object v0
+                        if-eqz v0, :continue_0
+                        invoke-virtual {v0}, Lde/simon/openinghours/models/OpeningHours;->getWeekdayText()Lio/realm/RealmList;
+                        move-result-object v0
+                        if-eqz v0, :continue_0
+                        check-cast v0, Ljava/util/List;
+                        invoke-direct {p0, v0}, Lde/simon/openinghours/views/custom/PlaceCard;->setWeekDayText(Ljava/util/List;)V
+
+                """,
+                continueLabel1,
```
oSumAtrIX, line 66: Please add comments on whats this doing

src/main/kotlin/app/revanced/patches/openinghours/crash/CrashPatch.kt
```diff
@@ -0,0 +1,86 @@
...
+                    move-result-object p1
+                    if-eqz p1, :continue_1
+                    invoke-virtual {p1}, Lde/simon/openinghours/models/OpeningHours;->getPeriods()Lio/realm/RealmList;
+                    move-result-object p1
+                    if-eqz p1, :continue_1
+                    check-cast p1, Ljava/util/List;
+                    invoke-direct {p0, p1}, Lde/simon/openinghours/views/custom/PlaceCard;->startCalculateStatus(Ljava/util/List;)V
+                """,
+                continueLabel2,
+            )
```
oSumAtrIX, line 48: Please add comments on whats this doing

oSumAtrIX (changes requested):
There's a lot going on in this patch. Please explain why this many changes are needed.

src/main/kotlin/app/revanced/patches/openinghours/crash/CrashPatch.kt
Linus789: Done :)

src/main/kotlin/app/revanced/patches/openinghours/crash/CrashPatch.kt
```diff
@@ -0,0 +1,94 @@
...
+
+            val continueLabel1 = ExternalLabel("continue_0", mutableMethod.getInstruction(setWeekDayTextIndex + 1))
+
+            // These instructions happen before the above instructions, but because of the
+            // changing indices for the instructions it was important to have the later
+            // instructions modified first.
+            //
+            // Again, we only added null checks and jump to label continue_0 if they return true.
```
oSumAtrIX, line 59: Can't you use replaceInstruction? Why do the original instructions crash?

Linus789: The original instructions crash because they look like this
```
invoke-virtual {p1}, Lde/simon/openinghours/models/Place;->getOpeningHours()Lde/simon/openinghours/models/OpeningHours;
move-result-object v0
invoke-static {v0}, Lkotlin/jvm/internal/Intrinsics;->checkNotNull(Ljava/lang/Object;)V
invoke-virtual {v0}, Lde/simon/openinghours/models/OpeningHours;->getWeekdayText()Lio/realm/RealmList;
move-result-object v0
invoke-static {v0}, Lkotlin/jvm/internal/Intrinsics;->checkNotNull(Ljava/lang/Object;)V
check-cast v0, Ljava/util/List;
invoke-direct {p0, v0}, Lde/simon/openinghours/views/custom/PlaceCard;->setWeekDayText(Ljava/util/List;)V
```

Yes, it should be possible to replace `Intrinsics;->checkNotNull` with the `if-eqz` check. I don't remember why I didn't do that in the first place. It seems like the only way adding new labels to a method is using `addInstructionsWithLabels`, though.

oSumAtrIX: You can create labels manually. Check how `addInstructionsWithLabels` is implemented. Also document things like this in the code

Linus789: Better like this?

src/main/kotlin/app/revanced/patches/openinghours/misc/fix/crash/FixCrashPatch.kt
```diff
@@ -0,0 +1,105 @@
...
+                indexedInstructions,
+                "Lde/simon/openinghours/views/custom/PlaceCard;",
+                "startCalculateStatus",
+            )
+
+            // Replace the Intrinsics;->checkNotNull instructions with a null check
+            // and jump to our newly created label if it returns true.
+            // This avoids throwing a NullPointerExceptions.
+            avoidNullPointerException(getOpeningHoursIndex[0], startCalculateStatusIndex, 4)
+            avoidNullPointerException(getOpeningHoursIndex[1], setWeekDayTextIndex, 0)
```
oSumAtrIX, line 74: The registers could change in future versions. Regardless of the version this patch is pinned to, it is better practice to get the register dynamically in case we want to remove the pin.

src/main/kotlin/app/revanced/patches/openinghours/misc/fix/crash/FixCrashPatch.kt
```diff
@@ -0,0 +1,105 @@
...
+@Suppress("unused")
+object FixCrashPatch : BytecodePatch(
+    setOf(SetPlaceFingerprint),
+) {
+
+    override fun execute(context: BytecodeContext) {
+        SetPlaceFingerprint.result?.let {
+            val indexedInstructions = it.mutableMethod.getInstructions().withIndex().toList()
+
+            fun avoidNullPointerException(openingHoursIndex: Int, jumpOverIndex: Int, targetRegister: Int) {
```
oSumAtrIX, line 31: Please give the parameters more concrete names. Currently, they are very generic, requiring you to infer what they are.

I have inferred the current names.

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/openinghours/misc/fix/crash/FixCrashPatch.kt
Linus789: Thanks for the refactor. :+1: I have now added the dynamic register use.

oSumAtrIX (approved):

PR #2708, opened 2024-02-09
feat(YouTube Music): Add `Settings` patch

oSumAtrIX:
# About

This PR adds a patch that adds ReVanced settings to YouTube Music.
This is achieved through recent abstractions regarding settings and resource patches.

The PR works like this:

Add a new settings patch, similar to YouTube, to YouTube Music. It extends existing abstract classes.
Because `BaseSettings` in integrations has to debug strings by default, the Debugging patch was abstracted in a similar fashion to a `BaseDebuggingPatch`. YouTube adds an additional debugging preference (Litho buffer logging), which does not exist in YouTube Music yet. Refer to the current implementation.

# Todo

- [ ] Properly `SettingsPatch` to `BaseSettingsPatch` similar to `BaseSettingsResourcePatch`
- [ ] Abstract "`ThemeHelper`" which is currently responsible in YouTube for selecting the correct theme for the settings screens
- [ ] Implement `ThemeHelper` in YouTube Music
- [ ] `ThemeHelper` may be merged into `SettingsPatch` in some way or another; it is exclusively used for that purpose, IIRC
- [ ] Resolve API breaking changes (Unlikely possible, but I really don't want to bump major again)
- [ ] IDE has messed up indentations in `strings.xml`, maybe fix them, but IDEs have the option to hide such changes in the diff viewer:
   image
- [ ] Implement options for the patches currently existing in YouTube Music

### DIFF
--- a/api/revanced-patches.api
+++ b/api/revanced-patches.api
@@ -346,6 +346,10 @@ public final class app/revanced/patches/music/misc/androidauto/BypassCertificate
 	public synthetic fun execute (Lapp/revanced/patcher/data/Context;)V
 }
 
+public final class app/revanced/patches/music/misc/debugging/DebuggingPatch : app/revanced/patches/shared/misc/debugging/BaseDebuggingPatch {
+	public static final field INSTANCE Lapp/revanced/patches/music/misc/debugging/DebuggingPatch;
+}
+
 public final class app/revanced/patches/music/misc/gms/Constants {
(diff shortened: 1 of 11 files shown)


YT-Advanced:
In YTM v6.43+, Google removed FullStackTraceActivity

oSumAtrIX:
That could work. What about apps for older Android versions?

Codrax:
Any updates on this pull request? It would be awesome to finally get a settings menu to YTM like on the regular YouTube client, to make them on par.

PR #2730, opened 2024-02-14
feat(X): Add `Open links as query` patch

dic1911:
As title, let me know if I missed anything.


src/main/kotlin/app/revanced/patches/twitter/misc/browserchooser/BrowserChooserPatch.kt
```diff
@@ -0,0 +1,76 @@
...
+            )
+        })
+
+        val inject = """
+            invoke-static {p0, p1}, $className->$helperMethodName(Landroid/content/Context;Landroid/content/Intent;)$returnType
+            return-void
+        """.trimIndent()
+
+        val instructions = method.getInstructions()
+        result.mutableMethod.removeInstructions(0, instructions.count())
```
oSumAtrIX, line 72: Simply injecting at index 0 should suffice

oSumAtrIX (changes requested):
Looks good all in all. I'll push some small changes, until then you can look into the current review

src/main/kotlin/app/revanced/patches/twitter/misc/browserchooser/BrowserChooserPatch.kt
```diff
@@ -0,0 +1,76 @@
...
+            className,
+            helperMethodName,
+            listOf(ImmutableMethodParameter("Landroid/content/Context;", null, "context"), ImmutableMethodParameter("Landroid/content/Intent;", null, "intent")),
+            returnType,
+            AccessFlags.STATIC.value,
+            null,
+            null,
+            MutableMethodImplementation(5)
+        ).toMutable().apply {
+            addInstructions(
```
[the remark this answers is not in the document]
dic1911: Agreed, here's the PR for that.
https://github.com/ReVanced/revanced-integrations/pull/570

src/main/kotlin/app/revanced/patches/twitter/misc/browserchooser/BrowserChooserPatch.kt
dic1911: Just force pushed with things cleaned up, tested on my side and seems to be working good.

oSumAtrIX (changes requested):
Pushed some changes, please test them. (You have to include the patch as `use = false`).
In the integrations I have also removed instantiating a new intent and instead overwrote the action. Is that even necessary?

dic1911:
@oSumAtrIX the `--include`/`-i` flag doesn't work on my side, I can't test it with `use` set to false.

src/main/kotlin/app/revanced/patches/twitter/misc/links/OpenLinksWithAppChooserPatch.kt
```diff
@@ -0,0 +1,34 @@
...
+    use = false,
+)
+@Suppress("unused")
+object OpenLinksWithAppChooserPatch : BytecodePatch(
+    setOf(OpenLinkFingerprint),
+) {
+    private const val METHOD_REFERENCE =
+        "Lapp/revanced/integrations/twitter/patches/links/OpenLinksWithAppChooserPatch;->" +
+            "openWithChooser(Landroid/content/Context;Landroid/content/Intent;)V"
+    override fun execute(context: BytecodeContext) {
```
oSumAtrIX, line 25:
```suggestion

    override fun execute(context: BytecodeContext) {
```

PR #2737, opened 2024-02-17
feat(YouTube): Reorganize settings menu

LisoUseInAIKyrios:
This PR reorganizes the YouTube ReVanced settings.

### Changes made:
- Breaks apart the huge `Layout` menu into 3 separate menus:  `Feed`, `Player`, `General layout`
- Moves `Shorts`, `Seekbar`, `Swipe controls`, `Alternative thumbnails`, `Return You Tube Dislike`, `SponsorBlock` up to top level menus under `ReVanced Settings`
- Removes the horizontal separators from the settings
- Removes the summary text of the root `ReVanced` menu and the first level of sub menus (`Player`, `Seekbar`, etc)
- Moves the `ReVanced Settings` menu and all sub-menus to the top of each preference list (and not the bottom)

### Screenshots:
[Root menu](https://github.com/ReVanced/revanced-patches/assets/118716522/65a1cb37-1bc9-4ec3-b45d-cb9a121cb0cb)
[Feed](https://github.com/ReVanced/revanced-patches/assets/118716522/9740cddd-1e62-472b-899e-5f6682c8d0a6)
[Player](https://github.com/ReVanced/revanced-patches/assets/118716522/c8be94e0-bab3-467b-a61a-cd1943a200d7)
[General layout](https://github.com/ReVanced/revanced-patches/assets/118716522/79d7ee1e-d53d-4823-b4d5-ee0acf31bdd5)
[Seekbar](https://github.com/ReVanced/revanced-patches/assets/118716522/8be74ab6-a23f-4d5b-a2e1-ccedc58befb1)
[Misc](https://github.com/ReVanced/revanced-patches/assets/118716522/5ca42e33-8a6e-43af-af4b-428e03981fa1)

[Integration changes](https://github.com/ReVanced/revanced-integrations/pull/571)


ILoveOpenSourceApplications:
> The `Misc` menu does not contain anything layout related.

A small but odd request, as all the other root menu settings use full names and the only exception being Misc, can we have it changed to Miscellaneous?

ILoveOpenSourceApplications:
> [`Other layout`](https://github.com/ReVanced/revanced-patches/assets/118716522/93529418-bfcd-457e-b913-edda7fcc27cf)

 Also other layout sounds off, and a better term should be used. I would suggest `General layout` as in meaning things which are generally found within YouTube layout.  But it's a suggestion, and if a better name is found that can be used as well.

ILoveOpenSourceApplications:
> That sounds ok but my only gripe about 'Miscellaneous' is the word is large and visually off putting. 'Misc' is not rare to find in other apps either.

I would've agreed if there weren't settings such as `Return YouTube Dislike` and `Alternative thumbnail`. But compared to those `Miscellaneous` isn't looking that large.
> Maybe use 'Other' for this since Layout no longer uses the name, and it still describes it's everything else.

I still wouldn't go for others as it feels too vague. We can surely look for another word perhaps if that's the route preferred.

src/main/kotlin/app/revanced/patches/shared/misc/settings/preference/PreferenceScreen.kt
```diff
@@ -17,9 +19,38 @@ open class PreferenceScreen(
...
     summaryKey: String? = "${key}_summary",
+    sortStyle: SortStyle? = null,
     tag: String = "PreferenceScreen",
     val preferences: Set<BasePreference>
-) : BasePreference(key, titleKey, summaryKey, tag) {
+    // Alternatively instead of repurposing the group key for sorting, this could be done adding an extra
+    // bundle to the preference xml.  But doing that appears to require bundling and referencing
+    // an additional xml file or adding new attributes to the attrs.xml file.
+    // Since the group key is unused, for now repurposing the unused key is much simpler.
+) : BasePreference(if (sortStyle == null) key else (key + sortStyle.sortSuffix), titleKey, summaryKey, tag) {
```
oSumAtrIX, line 29: I am not sure if that is possible, but can preference XML nodes come with custom attributes that can be read at runtime? If so, the key wouldn't need to be abused. The reason I am raising concern for manipulating the key like this behind the scenes is that while you provide the parameter to specify a key for the preference screen, in the end it does not turn out to be the final key. Instead the key gets additionally transformed with a sortStyle suffix.

src/main/kotlin/app/revanced/patches/shared/misc/settings/preference/PreferenceScreen.kt
```diff
@@ -9,6 +10,7 @@ import org.w3c.dom.Document
  * @param key The key of the preference. If null, other parameters must be specified.
  * @param titleKey The key of the preference title.
  * @param summaryKey The key of the preference summary.
+ * @param sortStyle Sorting style to use. A null value gives [SortStyle.TITLE].
```
oSumAtrIX, line 13: What is the reason this is nullable? The way it is now, the key without a sort suffix, or the key with the title sort suffix will both result in the same sorting.

src/main/kotlin/app/revanced/patches/shared/misc/settings/preference/PreferenceScreen.kt
```diff
@@ -17,9 +19,38 @@ open class PreferenceScreen(
...
     val preferences: Set<BasePreference>
-) : BasePreference(key, titleKey, summaryKey, tag) {
+    // Alternatively instead of repurposing the group key for sorting, this could be done adding an extra
+    // bundle to the preference xml.  But doing that appears to require bundling and referencing
+    // an additional xml file or adding new attributes to the attrs.xml file.
+    // Since the group key is unused, for now repurposing the unused key is much simpler.
+) : BasePreference(if (sortStyle == null) key else (key + sortStyle.sortSuffix), titleKey, summaryKey, tag) {
+
+    /**
+     * Specifies how Integrations should sort a PreferenceScreen or PreferenceGroup.
```
oSumAtrIX, line 32: What exactly is meant with "PreferenceGroup" here? Currently, we nest preferences in child preference screens.

src/main/kotlin/app/revanced/patches/youtube/misc/settings/SettingsPatch.kt
```diff
@@ -104,11 +105,32 @@ object SettingsPatch : BytecodePatch(
...
+        // This is done because sorting by title scatters related items apart,
+        // and there is no other way to specify an ordering here without refactoring other code.
+        // Each screen is bundled only if one or more preference is added during patching.
+        val ADS = Screen("revanced_settings_screen_01",
+            "revanced_ads_screen_title", null)
+        // Alternative thumbnails is item 2
+        val LAYOUT_FEED = Screen("revanced_settings_screen_03",
+            "revanced_layout_feed_screen_title", null)
+        val LAYOUT_PLAYER = Screen("revanced_settings_screen_04",
+            "revanced_layout_player_screen_title", null)
```
oSumAtrIX, line 118: Whats the reason the title has is being specified here? The title key would be infered from the preference key already.

src/main/kotlin/app/revanced/patches/youtube/misc/settings/SettingsPatch.kt
```diff
@@ -104,11 +105,32 @@ object SettingsPatch : BytecodePatch(
...
-        val LAYOUT = Screen("revanced_layout_screen")
-        val VIDEO = Screen("revanced_video_screen")
-        val MISC = Screen("revanced_misc_screen")
+        // Screens are sorted in the UI by the screen key.
+        // This is done because sorting by title scatters related items apart,
+        // and there is no other way to specify an ordering here without refactoring other code.
+        // Each screen is bundled only if one or more preference is added during patching.
+        val ADS = Screen("revanced_settings_screen_01",
+            "revanced_ads_screen_title", null)
+        // Alternative thumbnails is item 2
```
oSumAtrIX, line 114: Whats the reason for the alternative thumbnails screen to be missing here? Is it because of the screen being a "Custom" screen launched via an intent? I have been considering to refactor them to regular screens through the API we have here. This would reduce two ways to create settings screens, to one, simplifying the API.

src/main/kotlin/app/revanced/patches/youtube/misc/settings/SettingsPatch.kt
```diff
@@ -104,11 +105,32 @@ object SettingsPatch : BytecodePatch(
     }
 
     object PreferenceScreen : BasePreferenceScreen() {
-        val ADS = Screen("revanced_ads_screen")
-        val INTERACTIONS = Screen("revanced_interactions_screen")
-        val LAYOUT = Screen("revanced_layout_screen")
-        val VIDEO = Screen("revanced_video_screen")
-        val MISC = Screen("revanced_misc_screen")
+        // Screens are sorted in the UI by the screen key.
```
oSumAtrIX, line 108: In the inline docs for PreferenceScreen, it is said, if sortStyle is null, the preference will be sorted by title. Here it is said, it will be sorted by key.

oSumAtrIX:
Once the PR is in main, an announcement can be made in the app that the ReVanced submenu in the settings has moved to the top.

src/main/kotlin/app/revanced/patches/youtube/layout/theme/ThemeBytecodePatch.kt
```diff
@@ -100,7 +100,9 @@ object ThemeBytecodePatch : BytecodePatch(
     override fun execute(context: BytecodeContext) {
         AddResourcesPatch(this::class)
 
-        SettingsPatch.PreferenceScreen.LAYOUT.addPreferences(SwitchPreference("revanced_gradient_loading_screen"))
+        SettingsPatch.PreferenceScreen.LAYOUT_FEED.addPreferences(
```
oSumAtrIX, line 103: This probably should be in LAYOUT_GENERAL

src/main/kotlin/app/revanced/patches/youtube/misc/settings/SettingsPatch.kt
oSumAtrIX: I think translations would be unaffected by the key. Translators focus more on the text than the keys used for the text and because it is more consistent to rely on the convention that the title key is inferred from the preference key, I think it's better to infere it

oSumAtrIX: The full constructor can be used by setting the summary to null, whereas the title would be inferred from the preference key. What is the reason that sorting is done by key? It looks like, the preferences are ordered in English starting with A(ds). Wouldn't it make sense to order by title then? 

Additionally the keys have the same name apart from the suffix. I believe the reason is because of the convention to prefix them with revanced_, because otherwise they could have a descriptive name, just like the title and be prefixed with a number to retain the current order instead of the suffix. An intermediate solution would be to add the numbering after revanced_.

n1cKz1:
Shouldn't grey separator be under general layout instead of player?

oSumAtrIX:
I thought about the same, I was also thinking moving them to a Layout submenu would declutter the current view, and you'd have one view where everything is regarding layouts, the minor drawback would be one more nesting.

ILoveOpenSourceApplications:
> Maybe should rename `Feed layout` and `Player layout` to simply `Feed` and `Player`.

Then why not change `General layout` to `General` while we're at it?

oSumAtrIX:
I'll review it once I have time. I'll merge before next release

oSumAtrIX (approved):

PR #2776, opened 2024-02-27
fix(YouTube - Spoof app version): Remove broken versions

oSumAtrIX:
YouTube is disabling versions below v18 now and spoofing to these versions will break the app.

Maybe integrations can bump to newer versions, if currently users have set it below v18

### DIFF
--- a/src/main/resources/addresources/values/arrays.xml
+++ b/src/main/resources/addresources/values/arrays.xml
@@ -5,17 +5,11 @@
                 <item>@string/revanced_spoof_app_version_target_entry_1</item>
                 <item>@string/revanced_spoof_app_version_target_entry_2</item>
                 <item>@string/revanced_spoof_app_version_target_entry_3</item>
-                <item>@string/revanced_spoof_app_version_target_entry_4</item>
-                <item>@string/revanced_spoof_app_version_target_entry_5</item>
-                <item>@string/revanced_spoof_app_version_target_entry_6</item>
             </string-array>
             <string-array name="revanced_spoof_app_version_target_entry_values">
                 <item>18.33.40</item>
                 <item>18.20.39</item>
                 <item>18.09.39</item>
-                <item>17.08.35</item>
-                <item>16.08.35</item>
-                <item>16.01.35</item>
             </string-array>
         </patch>
         <patch id="layout.startpage.ChangeStartPagePatch">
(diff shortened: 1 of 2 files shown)


oSumAtrIX:
@LisoUseInAIKyrios The API below 18 is being shut down for users since 2 weeks now. The issue is being reported on r/revancedapp for reference.v18 supports at min Android 8, so YouTube should stop working on Android 7 and below

Sam101011:
As the server error reached me today, I did some testing on RVX(yes osum, I know it's not revanced related) and v17.34.36(the last version, compatible with Android 6&7) is the oldest one that still works(a.k.a @kitadai31 's fork is still fine).

So A5 is officially out of the game with A6&-7 barely afloat.

Imo, 17.34.36 could be added to the spoof version menu if it continues to be like that, but that's just a suggestion.

Sam101011:
Unfortunately the old layout under videos is dead, but what this version does is restore the old library layout:
Screenshot_20240229_151051_app rvx android youtube_1

I think the people like me, who stayed on the old layout until it died, may still prefer the old library layout.

A side note:the spoof app version menu on RVX looks like this:
Screenshot_20240229_151444_app rvx android youtube_1

You may add the version I suggested or the version under "Restore old playlist shelf" or any other version that does this, which you seem fit.

oSumAtrIX:
v17 is only partially shut down til a specific version. I don't think it is a good idea to support versions that will basically be obsolete in couple weeks.

kitadai31:
@Sam101011
> v17.34.36 is the oldest one that still works

No, I tested and v17.33.42 is the oldest one.

\>=17.33.xx: Worked
\<=17.32.xx: Got 400 error
(Currently, the 400 error is A/B test. But even if the 400 error does not occur, I could not log into my Google account.)

@LisoUseInAIKyrios
> Does 17.34.36 have any changes compared to spoofing to 18.09.39? The UI appears to be the same, and it's not the old layout like 17.08.35 was.

Just FYI, the under the videos layout of 17.34.36 is not as old as 17.08.35, but it is a little older than the 18.09.39's one.

[Spoofed_to_17.39.35.png](https://github.com/ReVanced/revanced-patches/assets/90122968/98582e74-90e1-438a-9bfc-510c4f458599)
[Spoofed_to_17.38.35.png](https://github.com/ReVanced/revanced-patches/assets/90122968/e42706d0-3b7f-49a1-9ae9-d1a8e5e87c3e)

(Also library layouts.)

@oSumAtrIX
> v17 is only partially shut down til a specific version. I don't think it is a good idea to support versions that will basically be obsolete in couple weeks.

I don't think so...
Google typically does minimum version bumping like this about once a year. So I believe 17.33.xx+ will work for another 6-12 months.

oSumAtrIX:
@kitadai31 Please open a PR

Sam101011:
@kitadai31 you're correct about the oldest working version, my bad. Tbh youtube really surprised me by scrapping not also v16(which was expected) but also more than half of v17.

And good on you for spotting the differences between the 2 bottom player layouts. I'd have never done it with my incredible ability to not see stuff that's right in front of my eyes.

PR #2924, opened 2024-03-27
fix(YouTube Music): Fix compatibility with latest versions

oSumAtrIX:
# About

The codecs unlock patch has been removed because from what I am seeing, Opus 251 is now natively supported
The minimized playback and background playback has been merged to restore consistency between YouTube and Music patches. Perhaps they should split again at a later time.

Some patches have been renamed according to our conventions and patches have been refactored.

Most patches remain compatible with older versions, one or two don't anymore.

# Todo

- [x] Fix `Hide get premium`. It needs a little more changes because it seems not to hide reliably


nightznero:
Why YTM doesn't have suggested version like for e.x YT? Does that mean i can use the very last stable version without any issues?

oSumAtrIX:
Because all patches claim to work on any future version

PR #2952, opened 2024-03-29
feat(YouTube - GmsCore): Require ignoring battery optimizations

oSumAtrIX:
# About

This PR requests the user to disable or ignore battery optimizations for GmsCore. This is necessary to prevent playback from failing when GmsCore is killed in the background. It is also a suggested action by GmsCore itself.

One bug that I have been observing for a while is that, for some reason, the app is not correctly killed with `System.exit(0)`. This causes spam; for example, in WSA this launches the website multiple times repeatedly in an infinite loop. Offtopic though.

RE: https://github.com/ReVanced/revanced-integrations/pull/599

### DIFF
--- a/src/main/resources/addresources/values/strings.xml
+++ b/src/main/resources/addresources/values/strings.xml
@@ -14,7 +14,9 @@
         </patch>
         <patch id="misc.gms.BaseGmsCoreSupportResourcePatch">
             <string name="gms_core_not_installed_warning">GmsCore is not installed. Install it.</string>
-            <string name="gms_core_not_whitelisted_warning">Follow the \"Don\'t kill my app\" guide for GmsCore.</string>
+            <string name="gms_core_not_whitelisted_title">You need to change your device settings</string>
+            <string name="gms_core_not_whitelisted_using_battery_optimizations_message">Your GmsCore is not whitelisted from battery optimization.\n\nTo fix this, follow the \"Don\'t kill my app\" guide for GmsCore.</string>
+            <string name="gms_core_not_whitelisted_not_allowed_in_background_message">Your GmsCore is not allowed to run in the background.\n\nTo fix this, follow the \"Don\'t kill my app\" guide for GmsCore.</string>
         </patch>
     </app>
     <app id="youtube">
--- a/src/main/resources/addresources/values/strings.xml
+++ b/src/main/resources/addresources/values/strings.xml
@@ -15,9 +15,11 @@
         <patch id="misc.gms.BaseGmsCoreSupportResourcePatch">
             <string name="gms_core_dialog_title">Action needed</string>
             <string name="gms_core_dialog_ok_button_text">Open website</string>
-            <string name="gms_core_not_installed_message">GmsCore is not installed.\n\nTap to open the website to install.</string>
-            <string name="gms_core_not_whitelisted_using_battery_optimizations_message">GmsCore is not whitelisted for battery optimization.\n\nFollow the \"Don\'t kill my app\" guide for GmsCore.</string>
-            <string name="gms_core_not_whitelisted_not_allowed_in_background_message">GmsCore does not have permission to run in the background.\n\nFollow the \"Don\'t kill my app\" guide for GmsCore.</string>
+            <string name="gms_core_dialog_not_installed_message">GmsCore is not installed.\n\nTap to open the website to install.</string>
+            <string name="gms_core_dialog_not_whitelisted_using_battery_optimizations_message">GmsCore is not whitelisted for battery optimization.\n\nFollow the \"Don\'t kill my app\" guide for GmsCore.</string>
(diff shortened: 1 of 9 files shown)


src/main/resources/addresources/values/strings.xml
```diff
@@ -13,8 +13,8 @@
             <string name="revanced_settings_import_failure_parse">Import failed: %s</string>
         </patch>
         <patch id="misc.gms.BaseGmsCoreSupportResourcePatch">
-            <string name="gms_core_not_installed_warning">GmsCore is not installed. Please install.</string>
-            <string name="gms_core_not_running_warning">GmsCore is failing to run. Please follow the \"Don\'t kill my app\" guide for GmsCore.</string>
+            <string name="gms_core_not_installed_warning">GmsCore is not installed. Install it.</string>
+            <string name="gms_core_not_whitelisted_warning">GmsCore is not whitelisted from battery optimizations. Follow the \"Don\'t kill my app\" guide for GmsCore.</string>
```
oSumAtrIX, line 17: This toast may be too long, not sure how else to solve though, maybe a notification?

oSumAtrIX: I am not sure but I'll try adding a dialog there and see what it does

oSumAtrIX:
@LisoUseInAIKyrios The issue also occurs on device, the toast is spammed a couple of times, the site is only opened once though, indicating that something isn't properly killed. I think the activity is being attempted to launch multiple times. Online, I found some specific method being called in addition to System.exit to finish the current activity properly

src/main/resources/addresources/values/strings.xml
oSumAtrIX: Either that or the toast can be split into two as an intermediate solution

src/main/resources/addresources/values/strings.xml
```diff
@@ -13,8 +13,10 @@
             <string name="revanced_settings_import_failure_parse">Import failed: %s</string>
         </patch>
         <patch id="misc.gms.BaseGmsCoreSupportResourcePatch">
-            <string name="gms_core_not_installed_warning">GmsCore is not installed. Please install.</string>
-            <string name="gms_core_not_running_warning">GmsCore is failing to run. Please follow the \"Don\'t kill my app\" guide for GmsCore.</string>
+            <string name="gms_core_not_installed_warning">GmsCore is not installed. Install it.</string>
+            <string name="gms_core_not_whitelisted_title">Action needed</string>
+            <string name="gms_core_not_whitelisted_using_battery_optimizations_message">GmsCore is not whitelisted from battery optimization.\n\nFollow the \"Don\'t kill my app\" guide for GmsCore.\n\nClick \"Ok\" to open it.</string>
+            <string name="gms_core_not_whitelisted_not_allowed_in_background_message">GmsCore is can't run in the background.\n\nFollow the \"Don\'t kill my app\" guide for GmsCore.\n\nClick \"Ok\" to open it.</string>
```
mlnrDev, line 19:
```suggestion
            <string name="gms_core_not_whitelisted_not_allowed_in_background_message">GmsCore can't run in the background.\n\nFollow the \"Don\'t kill my app\" guide for GmsCore.\n\nClick \"Ok\" to open it.</string>
```

src/main/resources/addresources/values/strings.xml
```diff
@@ -13,8 +13,13 @@
             <string name="revanced_settings_import_failure_parse">Import failed: %s</string>
         </patch>
         <patch id="misc.gms.BaseGmsCoreSupportResourcePatch">
-            <string name="gms_core_not_installed_warning">GmsCore is not installed. Please install.</string>
-            <string name="gms_core_not_running_warning">GmsCore is failing to run. Please follow the \"Don\'t kill my app\" guide for GmsCore.</string>
+            <string name="gms_core_dialog_title">Action needed</string>
+            <string name="gms_core_dialog_ok_button_text">Open website</string>
+            <string name="gms_core_dialog_not_installed_message">GmsCore is not installed.\n\nTap to open the website to install.</string>
+            <string name="gms_core_dialog_not_whitelisted_using_battery_optimizations_message">GmsCore is not whitelisted for battery optimization.\n\nFollow the \"Don\'t kill my app\" guide for GmsCore.</string>
+            <string name="gms_core_dialog_not_whitelisted_not_allowed_in_background_message">GmsCore does not have permission to run in the background.\n\nFollow the \"Don\'t kill my app\" guide for GmsCore.</string>
```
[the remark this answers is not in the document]
oSumAtrIX: That's something GmsCore has to deal with. GmsCore was previously called MicroG so probably that's something they know about and may change in the future.

Jaakkonen:
Instead of the link to the guide on how to disable dozing/etc page which is lacking guides for example for Pixel phones, I'd think forwarding people to the microG self-check and click the battery optimization row in the bottom which launches an action to disable battery optimizations for that app even if the option is not shown in that specific phones settings app.

PR #2954, opened 2024-03-30
feat(Tumblr): Add `Fix old versions` patch

leumasme:
Recent changes in the tumblr API have made all old versions of the app break due to removal of Tumblr Live.
Many API requests fail with the error `Unknown attribute 'live_now' requested`.
API paths are defined via annotations. This patch stops old app versions from requesting the removed fields by patching the function to handle these annotations to remove occourances of `,?live_now`


src/main/kotlin/app/revanced/patches/tumblr/fixes/FixOldVersionsPatch.kt
```diff
@@ -0,0 +1,38 @@
...
+import app.revanced.patcher.extensions.InstructionExtensions.addInstructions
+import app.revanced.patcher.patch.BytecodePatch
+import app.revanced.patcher.patch.annotation.CompatiblePackage
+import app.revanced.patcher.patch.annotation.Patch
+import app.revanced.patches.tumblr.fixes.fingerprints.HttpPathParserFingerprint
+import app.revanced.util.exception
+
+@Patch(
+    name = "Fix old versions",
+    description = "Fixes old versions of the app to break due to Tumblr removing remnants of Tumblr Live from the API, " +
```
oSumAtrIX, line 13: Which version is last that worked? You can add that to the description to say "Prior version x".

src/main/kotlin/app/revanced/patches/tumblr/fixes/FixOldVersionsPatch.kt
```diff
@@ -0,0 +1,38 @@
...
+    setOf(HttpPathParserFingerprint),
+) {
+    override fun execute(context: BytecodeContext) =
+        HttpPathParserFingerprint.result?.let {
+            val endIndex = it.scanResult.patternScanResult!!.endIndex
+
+            it.mutableMethod.addInstructions(
+                endIndex + 1,
+                """
+                # p2 is "value" string which has the value of the annotation (= the request path)
```
oSumAtrIX, line 29: Please explain what "value" string means. Is it the request URL?

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/tumblr/fixes/FixOldVersionsPatch.kt
leumasme: It's the request path, as the comment says. "value" is the name of the parameter. The original source code of the function is also linked in a comment at the fingerprint definition
https://github.com/square/retrofit/blob/ebf87b10997e2136af4d335276fa950221852c64/retrofit/src/main/java/retrofit2/RequestFactory.java#L270-L302

oSumAtrIX: What 
```suggestion
                # Remove "?live_now" from the request path p2.
```

src/main/kotlin/app/revanced/patches/tumblr/fixes/FixOldVersionsPatch.kt
```diff
@@ -0,0 +1,38 @@
...
+import app.revanced.patcher.patch.annotation.Patch
+import app.revanced.patches.tumblr.fixes.fingerprints.HttpPathParserFingerprint
+import app.revanced.util.exception
+
+@Patch(
+    name = "Fix old versions",
+    description = "Fixes old versions of the app to break due to Tumblr removing remnants of Tumblr Live from the API, " +
+        "that cause many requests on old app versions to fail. This patch has no effect on newer versions of the app.",
+    compatiblePackages = [CompatiblePackage("com.tumblr")],
+    use = false,
```
oSumAtrIX, line 16: @leumasme The patch is disabled by default as its unsuitable for new versions, sounds good?

src/main/kotlin/app/revanced/patches/tumblr/fixes/FixOldVersionsPatch.kt
leumasme: Some painful binary search later, 33.3 is the first version that works; <=33.2 are broken. The current grammar of the description is also completely broken so i changed it, feel free to reword again.

src/main/kotlin/app/revanced/patches/tumblr/fixes/FixOldVersionsPatch.kt
leumasme: Yeah that's fine.

oSumAtrIX (approved):

PR #2955, opened 2024-03-30
fix(Tumblr): Restore compatibility with latest versions 

leumasme:
Due to more classes getting obfuscated in new app versions, two patches had to be fixed:
- fixed feature flags patch & fingerprint
- fixed gift message popup fingerprint

Due to the removal of Tumblr Live, the Disable Tumblr Live Patch is obsolete
- deprecated disable tumblr live patch


Tarmogos (commented):
[]()

src/main/kotlin/app/revanced/patches/tumblr/featureflags/fingerprints/GetFeatureValueFingerprint.kt
```diff
@@ -1,6 +1,8 @@
 package app.revanced.patches.tumblr.featureflags.fingerprints
```
Tarmogos, line 1: @Det ~~[]()~~

PR #2958, opened 2024-03-31
fix(YouTube - GmsCore support): Prompt to disable battery optimizations, if not done already

LisoUseInAIKyrios:
If GmsCore is not whitelisted from device power optimizations:
- Show a dialog explaining why they need to turn off power optimizations
- Open the prompt to turn off optimizations

This is much simpler than opening a website explaining more changes.

If GmsCore is not running in the background on app launch, then the dontkillmyapp.com is still opened, as that condition needs instructions specific to the users device and operating system (a simple exclude from battery optimizations likely will not fix it).

[Integration changes](https://github.com/ReVanced/revanced-integrations/pull/601)

<details>

dialog

permission

</details>


oSumAtrIX:
GmsCore is the new official name of MicroG. I don't think it makes sense to mislabel it. If at all, I'd call it MicroG GmsCore.

src/main/resources/addresources/values/strings.xml
```diff
@@ -13,14 +13,17 @@
             <string name="revanced_settings_import_failure_parse">Import failed: %s</string>
         </patch>
         <patch id="misc.gms.BaseGmsCoreSupportResourcePatch">
+            <string name="gms_core_dialog_title">Action needed</string>
+            <string name="gms_core_dialog_not_whitelisted_not_allowed_in_background_message">MicroG GmsCore does not have permission to run in the background.\n\nFollow the \"Don\'t kill my app\" guide for your phone, and apply the instructions to your MicroG installation.\n\nThis is required for this app to work.</string>
```
oSumAtrIX, line 17:
```suggestion
            <string name="gms_core_dialog_not_whitelisted_not_allowed_in_background_message">MicroG GmsCore does not have permission to run in the background.\n\nFollow the \"Don\'t kill my app\" guide for your phone, and apply the instructions to your MicroG installation.\n\nThis is required for the app to work.</string>
```

src/main/resources/addresources/values/strings.xml
```diff
@@ -13,14 +13,17 @@
...
+            <string name="gms_core_dialog_open_website_text">Open website</string>
+            <string name="gms_core_dialog_not_whitelisted_using_battery_optimizations_message">MicroG GmsCore battery optimizations must be turned off to prevent app crashes.\n\nPress the continue button, and allow turning off battery optimizations.</string>
```
oSumAtrIX, line 19: Turned off -> Disabled (Probably a simpler term)
App crashes -> Issues (Crashes arent the only issue, playback could also silently fail apparently)
```suggestion
            <string name="gms_core_dialog_not_whitelisted_using_battery_optimizations_message">MicroG GmsCore battery optimizations must be disabled to prevent issues.\n\nPress the continue button and disable battery optimizations.</string>
```

src/main/resources/addresources/values/strings.xml
```diff
@@ -13,14 +13,17 @@
...
-            <string name="gms_core_toast_not_whitelisted_message">Follow the \"Don\'t kill my app\" guide for GmsCore.</string>
-            <string name="gms_core_dialog_title">Action needed</string>
-            <string name="gms_core_dialog_not_whitelisted_using_battery_optimizations_message">GmsCore is not whitelisted from battery optimization.\n\nFollow the \"Don\'t kill my app\" guide for GmsCore.</string>
-            <string name="gms_core_dialog_not_whitelisted_not_allowed_in_background_message">GmsCore does not have permission to run in the background.\n\nFollow the \"Don\'t kill my app\" guide for GmsCore.</string>
-            <string name="gms_core_dialog_ok_button_text">Open website</string>
+            <string name="gms_core_toast_not_whitelisted_message">MicroG GmsCore requires turning off battery optimizations</string>
+            <!-- Translations of this should not be longer than what is here, otherwise the text can be clipped and not entirely shown. -->
+            <string name="gms_core_toast_user_did_not_whitelist_message">Battery optimizations are still on, and this app may not work</string>
```
oSumAtrIX, line 26:
```suggestion
            <string name="gms_core_toast_not_whitelisted_message">MicroG GmsCore requires disabling battery optimizations</string>
            <!-- Translations of this should not be longer than what is here, otherwise the text can be clipped and not entirely shown. -->
            <string name="gms_core_toast_user_did_not_whitelist_message">Battery optimizations are still enabled and can cause issues</string>
```

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/music/misc/gms/GmsCoreSupportPatch.kt
```diff
@@ -21,7 +20,7 @@ object GmsCoreSupportPatch : BaseGmsCoreSupportPatch(
         CastDynamiteModuleV2Fingerprint,
         CastContextFetchFingerprint,
     ),
-    mainActivityOnCreateFingerprint = ApplicationInitFingerprint,
```
oSumAtrIX, line 24: If

src/main/kotlin/app/revanced/patches/music/misc/gms/fingerprints/MusicActivityOnCreateFingerprint.kt
```diff
@@ -0,0 +1,11 @@
+package app.revanced.patches.music.misc.gms.fingerprints
+
+import app.revanced.patcher.fingerprint.MethodFingerprint
+
+internal object MusicActivityOnCreateFingerprint : MethodFingerprint(
```
oSumAtrIX, line 5: Since `ApplicationInitFingerprint` (for integrations currently) has been replaced with `MusicActivityOnCreateFingerprint`, perhaps `MusicActivityOnCreateFingerprint` can be used for integrations as well instead of `ApplicationInitFingerprint`? If so, the fingerprint can be moved to a shared location.

oSumAtrIX (approved):

src/main/resources/addresources/values/strings.xml
```diff
@@ -14,13 +14,12 @@
...
-            <string name="gms_core_dialog_not_whitelisted_using_battery_optimizations_message">GmsCore is not whitelisted from battery optimization.\n\nFollow the \"Don\'t kill my app\" guide for GmsCore.</string>
-            <string name="gms_core_dialog_not_whitelisted_not_allowed_in_background_message">GmsCore does not have permission to run in the background.\n\nFollow the \"Don\'t kill my app\" guide for GmsCore.</string>
-            <string name="gms_core_dialog_ok_button_text">Open website</string>
+            <string name="gms_core_dialog_not_whitelisted_not_allowed_in_background_message">MicroG GmsCore does not have permission to run in the background.\n\nFollow the \"Don\'t kill my app\" guide for your phone, and apply the instructions to your MicroG installation.\n\nThis is required for the app to work.</string>
+            <string name="gms_core_dialog_open_website_text">Open website</string>
+            <string name="gms_core_dialog_not_whitelisted_using_battery_optimizations_message">MicroG GmsCore battery optimizations must be disabled to prevent issues.\n\nPress the continue button and disable battery optimizations.</string>
```
oSumAtrIX, line 21:
```suggestion
            <string name="gms_core_dialog_not_whitelisted_using_battery_optimizations_message">MicroG GmsCore battery optimizations must be disabled to prevent issues.\n\nTap on the continue button and disable battery optimizations.</string>
```

d-e-n-n-i-s-t:
Why does this have to keep annoying users that don't need to whitelist MicroG GmsCore? I don't use background playback/notifications, so why do I have to dismiss this annoying popup everytime I open ReVanced YouTube? Atleast create an option to disable this popup.

oSumAtrIX:
GmsCore must run in the background for the app to work correctly, not just for notifications.

d-e-n-n-i-s-t:
Not true from my experience, I never had any issues with it's Battery Optimization Setting being on restricted (worse than default, "optimized"). Weirdly, even background playback works for me up to an hour (haven't used it longer yet)

oSumAtrIX:
If it runs anyways in the background you have no reason to worry about setting it to unrestricted. Once you can provide evidence on setting the option to unrestricted being problematic we can reconsider.

d-e-n-n-i-s-t:
Yeah, Im not doing that for anyone to "reconsider". It's a matter of principle; only apps that absolutely need it, should be set to unrestricted - it doesn't need to run in idle sleep.
I guess this is where it stays then, just want to annotate that its really inconsiderate to not create a toggle for that popup. Seems to me like it goes against the general pathos of ReVanced, as you can turn off any annoying popups, except this one.

oSumAtrIX:
> Yeah, Im not doing that 

Feel free to open an issue once you change your mind, as it stands, no change is necessary.

PR #2971, opened 2024-04-01
feat(YouTube): Support version `19.09.38`, `19.10.39` and `19.11.43`

LisoUseInAIKyrios:


inotia00:
This is not because the `Swipe controls` patch is unstable.

The reason why stacktrace outputs the class `app.revanced.integrations.youtube.swipecontrols.SwipeControlsHostActivity` is simply because the `Swipe controls` patch [changes the SuperClass](https://github.com/ReVanced/revanced-patches/blob/c1c31540ccec9d146f411f1930bc95f9286a1a18/src/main/kotlin/app/revanced/patches/youtube/interaction/swipecontrols/SwipeControlsBytecodePatch.kt#L61) of the `MainActivity` class to the `SwipeControlsHostActivity` class. 

In other words, this is not an exception thrown by the `SwipeControlsHostActivity` class, it's just an exception in the `MainActivity` class.

Some notable points from this crash log include:
```
System.err: Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.Class java.lang.Object.getClass()' on a null object reference
System.err: at hhz.k(PG:1)
System.err: at kuz.G(PG:779)
```

This is the first exception thrown in the 'G' method of the 'kuz' class.

In YouTube v19.12.41, the 'G' method of the 'kuz' class handles the YouTube player button.

`Hide autoplay button` patch is probably the cause of the crash.

oSumAtrIX:
It is likely that some offsets have changed and the patch fails due to that. I remember it using offsets.

0xrxL:
> 19.13 is still beta, but it has even more changes and a few fingerprints no longer resolve.

No more... 19.13.36 was released into stable channel in April 3.

P.S: Anyway... autoplay button code is indeed changed starting from 19.13.xx (it doesn't support the old way to add it in layout anymore). I have already modified the old patch to make it compatible.

0xrxL:
> And with '19.13' the minimized playback code was changed.
>
> This PR is working ok with '19.11.43' and I've had no problems with it. Can modify this PR to add 19.10 and 19.11 and handle 19.13+ at a later time.

This Version (some weeks ago) broke all patches on my patcher, due to a weird debug lines adding by Google on many parts of the smali code (I'm talking about 3 debug lines per instruction).

Never seen something like that.

oSumAtrIX:
Sounds good, I'd like to review the PR first, bare with me

oSumAtrIX (approved):

PR #2988, opened 2024-04-03
fix(Tumblr - Fix old versions): Improve reliability by removing remnances of Tumblr Live 

leumasme:
The old `Fix old versions` patch still left some requests broken for two reasons:
- there was another broken field being requested: `?live_streaming_user_id`
- the requested fields were sometimes not specified directly in on the endpoint function (but instead supplied as a parameter with the @\Query annotation)

These issues are now resolved
- Patch now removes `,?live_streaming_user_id` from URLs as well
- added additional injection into retrofit2 [RequestBuilder.addQueryParam](https://github.com/square/retrofit/blob/trunk/retrofit/src/main/java/retrofit2/RequestBuilder.java#L186) to remove blocked fields even when not directly specified in the request path

### DIFF
--- a/api/revanced-patches.api
+++ b/api/revanced-patches.api
@@ -1062,6 +1062,10 @@ public final class app/revanced/patches/tumblr/fixes/FixOldVersionsPatch : app/r
 	public synthetic fun execute (Lapp/revanced/patcher/data/Context;)V
 }
 
+public final class app/revanced/patches/tumblr/fixes/fingerprints/AddQueryParamFingerprint : app/revanced/patcher/fingerprint/MethodFingerprint {
+	public static final field INSTANCE Lapp/revanced/patches/tumblr/fixes/fingerprints/AddQueryParamFingerprint;
+}
+
 public final class app/revanced/patches/tumblr/fixes/fingerprints/HttpPathParserFingerprint : app/revanced/patcher/fingerprint/MethodFingerprint {
 	public static final field INSTANCE Lapp/revanced/patches/tumblr/fixes/fingerprints/HttpPathParserFingerprint;
 }
--- a/api/revanced-patches.api
+++ b/api/revanced-patches.api
@@ -1062,14 +1062,6 @@ public final class app/revanced/patches/tumblr/fixes/FixOldVersionsPatch : app/r
 	public synthetic fun execute (Lapp/revanced/patcher/data/Context;)V
 }
 
-public final class app/revanced/patches/tumblr/fixes/fingerprints/AddQueryParamFingerprint : app/revanced/patcher/fingerprint/MethodFingerprint {
-	public static final field INSTANCE Lapp/revanced/patches/tumblr/fixes/fingerprints/AddQueryParamFingerprint;
-}
-
-public final class app/revanced/patches/tumblr/fixes/fingerprints/HttpPathParserFingerprint : app/revanced/patcher/fingerprint/MethodFingerprint {
-	public static final field INSTANCE Lapp/revanced/patches/tumblr/fixes/fingerprints/HttpPathParserFingerprint;
-}
-
 public final class app/revanced/patches/tumblr/live/DisableTumblrLivePatch : app/revanced/patcher/patch/BytecodePatch {
 	public static final field INSTANCE Lapp/revanced/patches/tumblr/live/DisableTumblrLivePatch;
 	public fun execute (Lapp/revanced/patcher/data/BytecodeContext;)V
--- a/src/main/kotlin/app/revanced/patches/tumblr/fixes/FixOldVersionsPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/tumblr/fixes/FixOldVersionsPatch.kt
@@ -28,7 +28,7 @@ object FixOldVersionsPatch : BytecodePatch(
 
         HttpPathParserFingerprint.result?.let {
             val endIndex = it.scanResult.patternScanResult!!.endIndex
-            // Remove the live query parameters from the path.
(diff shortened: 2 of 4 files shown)


api/revanced-patches.api
```diff
@@ -1062,10 +1062,6 @@ public final class app/revanced/patches/tumblr/fixes/FixOldVersionsPatch : app/r
 	public synthetic fun execute (Lapp/revanced/patcher/data/Context;)V
 }
 
-public final class app/revanced/patches/tumblr/fixes/fingerprints/HttpPathParserFingerprint : app/revanced/patcher/fingerprint/MethodFingerprint {
```
oSumAtrIX, line 1065: Fingerprint was supposed to be internal, API was never supposed to be present, treating as a fix.

src/main/kotlin/app/revanced/patches/tumblr/fixes/FixOldVersionsPatch.kt
```diff
@@ -5,34 +5,58 @@ import app.revanced.patcher.extensions.InstructionExtensions.addInstructions
...
+            }
         } ?: throw HttpPathParserFingerprint.exception
+
+        AddQueryParamFingerprint.result?.let {
+            // Remove the live query parameters when passed via the @Query annotation.
+            for (liveQueryParameter in liveQueryParameters) {
+                it.mutableMethod.addInstructions(
+                    0,
+                    """
+                    # queryParameterValue = queryParameterValue.replace(liveQueryParameter, "")
```
oSumAtrIX, line 52: What is "queryParameterValue"?

src/main/kotlin/app/revanced/patches/tumblr/fixes/FixOldVersionsPatch.kt
```diff
@@ -5,34 +5,58 @@ import app.revanced.patcher.extensions.InstructionExtensions.addInstructions
...
-                """
-                # Remove "?live_now" from the request path p2.
-                # p2 = p2.replace(p1, p3)
-                const-string p1, ",?live_now"
-                const-string p3, ""
-                invoke-virtual {p2, p1, p3}, Ljava/lang/String;->replace(Ljava/lang/CharSequence;Ljava/lang/CharSequence;)Ljava/lang/String;
-                move-result-object p2
-            """,
-            )
+            // Remove the live query parameters from the path.
```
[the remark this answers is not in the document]
leumasme: An example of an API path is:
`/blog/somename?fields[blog]=uuid,username,?is_admin,?live_now`
The fields list is here treated as a single query parameter with the name `fields[blog]`
and the value `uuid,username,?is_admin,?live_now`
We remove `,?live_now` from the value, so the new path becomes
`/blog/somename?fields[blog]=uuid,username,?is_admin,?live_now`
This theoretically doesn't work if ?live_now were the first field due to the leading comma, but that doesn't seem to be the case in any requests (and the parameter list requested is always quite long, >20)

src/main/kotlin/app/revanced/patches/tumblr/fixes/FixOldVersionsPatch.kt
leumasme: See other comment, ex:
path = `/blog/somename?fields[blog]=uuid,username,?is_admin,?live_now`
query parameter value = `uuid,username,?is_admin,?live_now`

src/main/kotlin/app/revanced/patches/tumblr/fixes/FixOldVersionsPatch.kt
```diff
@@ -21,41 +21,40 @@ object FixOldVersionsPatch : BytecodePatch(
...
-                """
+                """,
                 )
             }
         } ?: throw HttpPathParserFingerprint.exception
 
         AddQueryParamFingerprint.result?.let {
-            // Remove each of the blockedStrings elements when passed in as a @Query argument
-            for (blockedString in blockedStrings) {
+            // Remove the live query parameters when passed via the @Query annotation.
```
leumasme, line 47: Not sure if this rewording is accurate.
An argument (parameter) of the function is annotated with @\Query("query parameter name").
When the function is called, the value of the corresponding query parameter will be set by the value passed for the parameter.
e.g.
```java
    @GET("blog/{hostname}/info?is_full_blog_info=false")
    Call<ApiResponse<BlogInfoResponse>> getBlogInfoPartial(@Path("hostname") String str, @Query("blog_name") String str2, @Query("prefetch[0]") String str3, @Query("fields[blogs]") String str4);
```
and then
```java
instance.getBlogInfoPartial(getHostname(), getBlogname(), "", "?can_be_booped,name,?live_now,...")
```
So the name of the query parameter is determined by the annotation; its value is determined by the function argument

oSumAtrIX: That makes sense, please adjust the rewording.

src/main/kotlin/app/revanced/patches/tumblr/fixes/FixOldVersionsPatch.kt
oSumAtrIX: Odd, thats invalid syntax. ? only denotes the start of the query parameters. The following query parameters are separated by &.

leumasme: The '?' field prefix here denotes the field being optional. I'm not sure if it's url encoded along the way or if their server software just ignores question marks past the first and accepts them as valid query parameter values, but the patch works.

oSumAtrIX (approved):

rizwan-r-r (approved):

PR #3000, opened 2024-04-07
feat(Amazon): Add `Always allow deep-linking` patch

1fexd:
Amazon's mobile app (`com.amazon.mShop.android.shopping`) disables its deep-linking activity in certain cases (when it is not set to open its links by default), preventing it from being launched even via an explicit Intent.

This patch removes that restriction.

### DIFF
--- a/src/main/kotlin/app/revanced/patches/amazon/deeplinking/DeepLinkingPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/amazon/deeplinking/DeepLinkingPatch.kt
@@ -8,7 +8,7 @@ import app.revanced.patcher.patch.annotation.Patch
 
 @Patch(
     name = "Always allow deep-linking",
-    description = "Always allow deep-linking",
+    description = "Allows the Amazon app to open Amazon links, even if it is not the default link handler for Amazon domains.",
     compatiblePackages = [CompatiblePackage("com.amazon.mShop.android.shopping")]
 )
 @Suppress("unused")
--- a/src/main/kotlin/app/revanced/patches/amazon/deeplinking/DeepLinkingPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/amazon/deeplinking/DeepLinkingPatch.kt
@@ -8,7 +8,7 @@ import app.revanced.patcher.patch.annotation.Patch
 
 @Patch(
     name = "Always allow deep-linking",
-    description = "Allows the Amazon app to open Amazon links, even if it is not the default link handler for Amazon domains.",
+    description = "Open Amazon links, even if the app is not set to handle Amazon links.",
     compatiblePackages = [CompatiblePackage("com.amazon.mShop.android.shopping")]
 )
 @Suppress("unused")
--- a/src/main/kotlin/app/revanced/patches/amazon/deeplinking/DeepLinkingPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/amazon/deeplinking/DeepLinkingPatch.kt
@@ -1,0 +1,27 @@
+package app.revanced.patches.amazon.deeplinking
+
+import app.revanced.patcher.data.BytecodeContext
+import app.revanced.patcher.extensions.InstructionExtensions.addInstructions
+import app.revanced.patcher.patch.BytecodePatch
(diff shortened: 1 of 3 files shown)


src/main/kotlin/app/revanced/patches/amazon/deeplinking/DeepLinkingPatch.kt
```diff
@@ -0,0 +1,27 @@
...
+
+import app.revanced.patcher.data.BytecodeContext
+import app.revanced.patcher.extensions.InstructionExtensions.addInstructions
+import app.revanced.patcher.patch.BytecodePatch
+import app.revanced.patcher.patch.annotation.CompatiblePackage
+import app.revanced.patcher.patch.annotation.Patch
+
+@Patch(
+    name = "Always allow deep-linking",
+    description = "Always allow deep-linking",
```
oSumAtrIX, line 11: It would be nice to use the description to clarify what deep linking is/what the patch is useful for.

src/main/kotlin/app/revanced/patches/amazon/deeplinking/DeepLinkingPatch.kt
```diff
@@ -0,0 +1,27 @@
...
+    compatiblePackages = [CompatiblePackage("com.amazon.mShop.android.shopping")]
```
oSumAtrIX, line 12: Just to be sure, the package name casing is intended here?

src/main/kotlin/app/revanced/patches/amazon/deeplinking/DeepLinkingPatch.kt
1fexd: Done

src/main/kotlin/app/revanced/patches/amazon/deeplinking/DeepLinkingPatch.kt
1fexd: Yes

src/main/kotlin/app/revanced/patches/amazon/deeplinking/DeepLinkingPatch.kt
```diff
@@ -0,0 +1,27 @@
...
+
+import app.revanced.patcher.data.BytecodeContext
+import app.revanced.patcher.extensions.InstructionExtensions.addInstructions
+import app.revanced.patcher.patch.BytecodePatch
+import app.revanced.patcher.patch.annotation.CompatiblePackage
+import app.revanced.patcher.patch.annotation.Patch
+
+@Patch(
+    name = "Always allow deep-linking",
+    description = "Allows the Amazon app to open Amazon links, even if it is not the default link handler for Amazon domains.",
```
oSumAtrIX, line 11: Did I understood correctly:

```suggestion
    description = "Open Amazon links, even if the app is not set to handle Amazon links.",
```

oSumAtrIX: If so, why is the patch necessary? Couldn't you set it as the handler for Amazon links?

1fexd: Amazon is special in that regard, as they query the package manager with a VIEW Intent to check the installed handlers for the currently selected marketplace region (e.g. `https://www.amazon.de).

It appears that due to the domain verification system for deep links introduced in Android 12, this query only returns a result if the link handler for the queried domain has cryptographically proven that they possess ownership over that domain, and is enabled as a link handler.

If no results are returned from the query (either because the user specifically toggled link handling off for Amazon, or if the system couldn't resolve the domain verification status), Amazon then just goes ahead and disables the Activity (via `PackageManager#setComponentEnabledSetting`). This means that even if you launch an explicit Intent which specifies exactly this Activity/Component, it can no longer be launched. 

This behavior is implemented in `com.amazon.mShop.httpUrlDeepLink.DeepLinkingModule`:

image

The method which this patch patches to always return true:

image

This is the method which disables the Activity/Component:

image

oSumAtrIX (approved):

PR #3044, opened 2024-04-18
feat(Spotify): Add `Unlock Spotify Premium` patch

oSumAtrIX:
# About

This PR adds a patch for Spotify to enable Spotify Premium by patching a native binary using a raw resource patch.
The question would be whether this patch should be merged, given that it's a patch to unlock Spotify Premium.

### DIFF
--- a/api/revanced-patches.api
+++ b/api/revanced-patches.api
@@ -965,6 +965,10 @@ public final class app/revanced/patches/spotify/navbar/PremiumNavbarTabResourceP
 	public fun execute (Lapp/revanced/patcher/data/ResourceContext;)V
 }
 
+public final class app/revanced/patches/spotify/premium/UnlockPremiumPatch : app/revanced/patches/shared/misc/hex/BaseHexPatch {
+	public fun <init> ()V
+}
+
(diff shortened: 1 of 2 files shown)


oSumAtrIX:
xManager is closed source and also muddies the patched apps; ReVanced, on the other hand, is open source and providers transparency, apart from that feature wise they should not differ.

oSumAtrIX:
The contribution guidelines outline that patches that provide significant value are treated on a case-by-case basis, but I am also not entirely sure if we want a patch for an app that is known for [aggressively fighting things like that](https://www.reddit.com/r/spotify/comments/c5zpe3/hate_the_new_spotify_if_youre_on_android_i_have/). 

Users can technically use the standalone `Hex` patch and input the replacements themselves to achieve what this patch does manually.

Ushie:
xManager has existed for a decent amount of time and they directly distribute the modded versions over the internet, by now I assume they'd have been DMCA'd if Spotify cared, what's the worst-case scenario? we have to take down the patch?

oSumAtrIX:
I think it's fine to not merge this PR. At least it can act as a PoC and others can take an example from it. There are people that pick up dropped patches and eventually when ReVanced Manager allows, those can be used side by side with other patches such as ours. xManager technically bypasses DMCAs since it merely downloads the modded apps, unlike us.

PR #3109, opened 2024-05-02
feat(Photomath): Support version `8.37.0`

FullerBread2032:
Fixed CheckSignatureFingerprint and GetDeviceIdFingerprint 
Changed supported version from `8.32.0` to `8.37.0`

### DIFF
--- a/src/main/kotlin/app/revanced/patches/photomath/detection/deviceid/fingerprints/GetDeviceIdFingerprint.kt
+++ b/src/main/kotlin/app/revanced/patches/photomath/detection/deviceid/fingerprints/GetDeviceIdFingerprint.kt
@@ -1,9 +1,21 @@
 package app.revanced.patches.photomath.detection.deviceid.fingerprints
 
 import app.revanced.patcher.fingerprint.MethodFingerprint
+import com.android.tools.smali.dexlib2.Opcode
 
 internal object GetDeviceIdFingerprint : MethodFingerprint(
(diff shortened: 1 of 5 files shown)


patches.json
oSumAtrIX, line 1: This file is autogenerated and should not be committed to

src/main/kotlin/app/revanced/patches/photomath/detection/deviceid/fingerprints/GetDeviceIdFingerprint.kt
```diff
@@ -1,9 +1,21 @@
...
+        Opcode.IGET_OBJECT,
+        Opcode.INVOKE_STATIC,
+        Opcode.MOVE_RESULT_OBJECT,
+        Opcode.IF_NEZ,
+        Opcode.INVOKE_STATIC,
+        Opcode.MOVE_RESULT_OBJECT,
+        Opcode.INVOKE_VIRTUAL,
+        Opcode.MOVE_RESULT_OBJECT,
+        Opcode.INVOKE_VIRTUAL,
+    ),
```
oSumAtrIX, line 19: Is there any other way than using an opcode pattern such as strings and access flags?

oSumAtrIX (changes requested):
This PR goes against contribution guidelines, but I'm making an exception for this

oSumAtrIX (approved):

PR #3110, opened 2024-05-02
feat(WarnWetter - Promo code unlock): Constrain to last working version

Bceez:


Technikte:
@Bceez It would be just better to remove "FirebaseGetCertPatch" and all dependencies because after version WarnWetter 5.0.1 -> "We do not collect userdata with Google Analytics for Firebase", so that means the firebase patch is not needed anymore and thedependented firebasepatch fails.

I tested locally and after removing dep the patch works on all versions after 5.0.1, tested up to newest 5.0.3.

Bceez:
@Technikte The reason for the version constrain is that this patch unlock premium features which are against the [vontribution guidelines](https://github.com/ReVanced/revanced-patches/blob/main/CONTRIBUTING.md)

Technikte:
> @Technikte The reason for the version constrain is that this patch unlock premium features which are against the [contribution guidelines](https://github.com/ReVanced/revanced-patches/blob/main/CONTRIBUTING.md)

Well, I see then the version constrain is the next best thing to do..

PR #3144, opened 2024-05-09
feat(Piccoma): Add `Enable debugging` patch

Starmania:


src/main/kotlin/app/revanced/patches/piccomafr/misc/EnableDebuggingPatch.kt
```diff
@@ -0,0 +1,79 @@
...
+import app.revanced.patches.piccomafr.misc.fingerprints.AppConfFingerprint
+import app.revanced.patches.piccomafr.misc.fingerprints.OpenToUserFingerprint
+import app.revanced.util.exception
+import com.android.tools.smali.dexlib2.AccessFlags
+import com.android.tools.smali.dexlib2.immutable.ImmutableField
+import com.android.tools.smali.dexlib2.immutable.value.ImmutableBooleanEncodedValue
+
+@Patch(
+    name = "Enable debugging",
+    description = "Enable every debug possibilities of the app.",
```
oSumAtrIX, line 20: Please provide a better description of what this patch does. What kind of debugging is achieved?

src/main/kotlin/app/revanced/patches/piccomafr/misc/EnableDebuggingPatch.kt
```diff
@@ -0,0 +1,79 @@
...
+        ),
+    ],
+    use = false,
+)
+@Suppress("unused")
+object EnableDebuggingPatch : BytecodePatch(
+    setOf(AppConfFingerprint, OpenToUserFingerprint),
+) {
+    override fun execute(context: BytecodeContext) {
+        AppConfFingerprint.result?.mutableMethod?.apply {
```
oSumAtrIX, line 49: Please add a comment here explaining what each patch does. It is unclear what the instruction replacement is done for and what it effects. Add comments for the patches below as well.

src/main/kotlin/app/revanced/patches/piccomafr/misc/EnableDebuggingPatch.kt
Starmania: I would like your opinion on it, but I think I will split the patch in 3. One part for DEV mode, one part for logging HTTP requests by setting the DEBUG flag in volley and a last one for the rest. What do you think ?

Starmania:
No, there's no point to finish it. The application will end on September 30 of this year and it wasn't ground breaking features. It would have been useful only if development had continued, as it would have enabled access to some beta/dev functionality.

PR #3154, opened 2024-05-11
feat(Boost For Reddit): Add `Fix /s/ links` patch

### LINKED ISSUE #478: feat: Boost for Reddit - opening new kind of Reddit URL in the shape of https://reddit.com/r/.../s/...
### Application

Boost

### Issue

There has been a new kind of Reddit URL, generated with the "Share" feature of the latest Reddit Mobile. It is in the shape of `https://reddit.com/r/<subreddit>/s/<Base 62 code>`, and might point to a subreddit, post, or comment. Boost cannot correctly open it. Instead Boost will only open the subreddit.

Example: URL 

https://reddit.com/r/badUIbattles/s/0iHSWsyEWu 

should lead to post 

https://www.reddit.com/r/badUIbattles/comments/15nbq9x/usernames_in_comments_are_sometimes_cut_off/

But instead opens https://www.reddit.com/r/badUIbattles/ in Boost.

### Patch

The patch should hopefully be able to correctly open the subreddit, post, or comment that this new kind of Reddit URL points to.

### Motivation

If Reddit Mobile causes this new type of URLs to gradually replace the old type, Boost users will find it hard to keep using only Boost for Reddit. This patch will keep Boost's usability as high as before.

### Acknowledgements

OctoNezd:
Multi-part PR:
- Part 2: https://github.com/ReVanced/revanced-integrations/pull/631

This PR fixes #478 

Code is essentially copy-paste from patch for Sync, but with more exception handling and adding the Authorization header to prevent reddit from giving the client page "log in or provide token or else" without proper Location: header.


OctoNezd:
Do I commit the dumped APIs in?

oSumAtrIX:
Yes

oSumAtrIX (changes requested):
Looks good. Something that should be done before merging would be to introduce a base class for the fix s link patch similar to the client spoof base class. The implementations are very similar and as such qualify for implementations of such a base class.

OctoNezd:
Ready for merge.

oSumAtrIX (commented):
I have pushed refactoring changes (also to integrations). Check if they look good. If so, the PR is ready to be merged.

OctoNezd:
Looks to me that patch is broken now: boost throws error and doesnt say anything in logs (cause timber debug isnt enabled, no idea where to enable it), and sync throws that in logs: `java.lang.NullPointerException: Attempt to invoke virtual method 'void app.revanced.integrations.shared.fixes.slink.BaseFixSLinksPatch.setAccessToken(java.lang.String)' on a null object reference`

OctoNezd:
Fixed that in https://github.com/ReVanced/revanced-integrations/pull/631/commits/85c2ebe3a83e79578643eb3879cab97b0055aa2a

Ready to merge.

PR #3156, opened 2024-05-12
feat(Reddit - Hide banner ads): Constrain to last working version

ascopes:
Until GH-3099 is fixable and addressed, this limits the version of Reddit that can be
used to 2024.0.17 which is the last known working version.

### DIFF
--- a/src/main/kotlin/app/revanced/patches/reddit/ad/banner/HideBannerPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/reddit/ad/banner/HideBannerPatch.kt
@@ -6,7 +6,7 @@ import app.revanced.patcher.patch.annotation.Patch
 
 @Patch(
     description = "Hides banner ads from comments on subreddits.",
-    // Pinned to 2024.0.17 until https://github.com/iBotPeaches/Apktool/issues/3534
+    // Constrained to last working version (2024.0.17) until https://github.com/iBotPeaches/Apktool/issues/3534
     // is addressed which is causing crashes during patching.
     // See https://github.com/ReVanced/revanced-patches/issues/3099.
     compatiblePackages = [CompatiblePackage("com.reddit.frontpage", ["2024.0.17"])]
--- a/src/main/kotlin/app/revanced/patches/reddit/ad/banner/HideBannerPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/reddit/ad/banner/HideBannerPatch.kt
@@ -4,7 +4,13 @@ import app.revanced.patcher.data.ResourceContext
 import app.revanced.patcher.patch.ResourcePatch
 import app.revanced.patcher.patch.annotation.Patch
 
-@Patch(description = "Hides banner ads from comments on subreddits.")
+@Patch(
+    description = "Hides banner ads from comments on subreddits.",
+    // Pinned to 2024.0.17 until https://github.com/iBotPeaches/Apktool/issues/3534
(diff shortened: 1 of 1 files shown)


ascopes:
Not part of this MR but possibly worth noting that the link to the docs explaining how to structure the commit naming seems to be broken, so the name of this commit is a bit of a stab in the dark on my behalf.

Hopefully this is correct.

src/main/kotlin/app/revanced/patches/reddit/ad/banner/HideBannerPatch.kt
```diff
@@ -4,7 +4,13 @@ import app.revanced.patcher.data.ResourceContext
 import app.revanced.patcher.patch.ResourcePatch
 import app.revanced.patcher.patch.annotation.Patch
 
-@Patch(description = "Hides banner ads from comments on subreddits.")
+@Patch(
+    description = "Hides banner ads from comments on subreddits.",
+    // Pinned to 2024.0.17 until https://github.com/iBotPeaches/Apktool/issues/3534
```
oSumAtrIX, line 9:
```suggestion
    // Constrained to last working version (2024.0.17) until https://github.com/iBotPeaches/Apktool/issues/3534
```

oSumAtrIX (changes requested):
Since this patch is a dependency and not usable in any frontend, it inherits the compatibility of the dependant patches. For that reason, the compatibility needs to be set on the patches which depend on this one. Alternatively, this patch can get a name & other patches can stop depend on it so that they can be used on later versions with this patch unselected.

ascopes:
If I go for the latter, is it just a case of setting the name attribute directly or do other patches need modifying as well?

Apologise if it is a silly question, somewhat new to how this works!

src/main/kotlin/app/revanced/patches/reddit/ad/banner/HideBannerPatch.kt
```diff
@@ -4,7 +4,13 @@ import app.revanced.patcher.data.ResourceContext
 import app.revanced.patcher.patch.ResourcePatch
 import app.revanced.patcher.patch.annotation.Patch
 
-@Patch(description = "Hides banner ads from comments on subreddits.")
+@Patch(
+    description = "Hides banner ads from comments on subreddits.",
+    // Constrained to last working version (2024.0.17) until https://github.com/iBotPeaches/Apktool/issues/3534
+    // is addressed which is causing crashes during patching.
+    // See https://github.com/ReVanced/revanced-patches/issues/3099.
+    compatiblePackages = [CompatiblePackage("com.reddit.frontpage", ["2024.0.17"])]
```
programminghoch10, line 12:
```suggestion
    // Constrained to last working version (2024.17.0) until https://github.com/iBotPeaches/Apktool/issues/3534
    // is addressed which is causing crashes during patching.
    // See https://github.com/ReVanced/revanced-patches/issues/3099.
    compatiblePackages = [CompatiblePackage("com.reddit.frontpage", ["2024.17.0"])]
```
Wasn't the last working version `2024.17.0` and not `2024.0.17`?

programminghoch10 (changes requested):

ericswpark:
@ascopes why was the head repository deleted? Did another PR supersede this one?

ascopes:
Ah whoops, I was removing an unused clone for another revanced project and accidentally deleted the wrong one... will reraise with the correct fixes as given above when I get some time!

ascopes:
Reopened at #3192

PR #3164, opened 2024-05-14
fix(YouTube Music): Make `Hide 'Get Music Premium' label` and `Remove upgrade button` compatible with latest version

ponces:
Weirdly enough, this was the fingerprint used before the latest PR fix. So we are in fact returning to an old fingerprint...


oSumAtrIX:
I had mentioned in the previous PR that the compiler can't decide between the two instructions so it's likely it'll go back to the other pattern once again

oSumAtrIX (approved):

PR #3188, opened 2024-05-18
fix(YouTube Music): set compatible version for patches

sha-265:
Some patches supports only version `6.47.53` of YT music, so `compatiblePackages` should be set


sha-265:
fix should be released before next major version update

oSumAtrIX:
The dev branch indeed has already fixed compatibility, thanks regardless.

PR #3192, opened 2024-05-19
fix(Reddit): Pin Reddit to 2024.17.0 for HideAdsPatch

ascopes:
Until https://github.com/ReVanced/revanced-patches/issues/3099 is fixable and addressed, this limits the version of Reddit that can be
used to 2024.0.17 which is the last known working version.

---

Reopened this from #3156 as the repo was accidentally closed due to an error on my side cleaning up unused forks on my account.

Added feedback from that PR to fix versions and apply the constraint to dependent patches that are named.


src/main/kotlin/app/revanced/patches/reddit/ad/banner/HideBannerPatch.kt
```diff
@@ -4,7 +4,13 @@ import app.revanced.patcher.data.ResourceContext
 import app.revanced.patcher.patch.ResourcePatch
 import app.revanced.patcher.patch.annotation.Patch
 
-@Patch(description = "Hides banner ads from comments on subreddits.")
+@Patch(
+    description = "Hides banner ads from comments on subreddits.",
+    // Constrained to last working version (2024.17.0) until https://github.com/iBotPeaches/Apktool/issues/3534
+    // is addressed which is causing crashes during patching.
+    // See https://github.com/ReVanced/revanced-patches/issues/3099.
+    compatiblePackages = [CompatiblePackage("com.reddit.frontpage", ["2024.17.0"])]
```
ascopes, line 12: Kept this in so it is clear that if anything depends on this in the future that it needs to have this constraint.

Assuming the dependency graph does not resolve these requirements recursively given previous PR feedback?

oSumAtrIX: The annotation should be removed and instead a comment can be kept. Patcher does not handle this annotation field but the application that runs the patcher.

oSumAtrIX, line 12:
```suggestion
// Last working version is 2024.17.0
// until https://github.com/iBotPeaches/Apktool/issues/3534 is addressed 
// which is causing crashes during patching.
// See https://github.com/ReVanced/revanced-patches/issues/3099.
@Patch(
    description = "Hides banner ads from comments on subreddits.",
```

oSumAtrIX (approved):

src/main/kotlin/app/revanced/patches/reddit/ad/general/HideAdsPatch.kt
```diff
@@ -20,7 +20,12 @@ import com.android.tools.smali.dexlib2.iface.reference.MethodReference
 @Patch(
     name = "Hide ads",
     dependencies = [HideBannerPatch::class, HideCommentAdsPatch::class],
-    compatiblePackages = [CompatiblePackage("com.reddit.frontpage")],
+    // Constrained to last working version (2024.17.0) until https://github.com/iBotPeaches/Apktool/issues/3534
+    // is addressed which is causing crashes during patching.
+    // See https://github.com/ReVanced/revanced-patches/issues/3099.
+    // This is constrained due to the dependency on HideBannerPatch which
+    // does not get exposed directly to users.
```
oSumAtrIX, line 27:
```suggestion
	// Note that for now, this patch and anything using it will only work on
	// Reddit 2024.17.0 or older. Newer versions will crash during patching.
	// See https://github.com/ReVanced/revanced-patches/issues/3099 
 	// and https://github.com/iBotPeaches/Apktool/issues/3534.
    // This constraint is necessary due to dependency on HideBannerPatch.
```

oSumAtrIX (approved):

PR #3197, opened 2024-05-20
feat(YouTube - Playback speed): Add speed dialog button to video player

MarcaDian:
Adds a simply speed button to the bottom of video player that called flyout custom speed dialog same as from flyout menu.

<details>

Зображення WhatsApp, дата_ 2024-05-20 о 23 28 57_a81b6303
</details>

Partially closes https://github.com/ReVanced/revanced-patches/issues/1620

integrations  https://github.com/ReVanced/revanced-integrations/pull/636

PR #3220, opened 2024-05-22
fix(YouTube - Client spoof): Spoof iOS client model to fix various side effects

oSumAtrIX:
RE: https://github.com/ReVanced/revanced-integrations/pull/638

This PR also changes the default client spoof to iOS, as the only side effect known now is that HDR is not working.


zainarbani:
Is user-agent also spoofed as iOS client?

oSumAtrIX:
No

Pamilg8:
> Playback speed is not fixed with this PR.
> 
> Speed menu can be fixed by improving pr #3197

This is a replacement button. Original youtube button.

zainarbani:
> No

Screenshot_2024-05-23-19-23-22-806_com rhmsoft edit pro
Screenshot_2024-05-23-19-23-15-319_com termux

Could possibly related?

oSumAtrIX:
Good find for HDR. It may bring back the playback speed menu entry given that iOS clients are supposed to have one to begin with.

zainarbani:
> Good find for HDR. It may bring back the playback speed menu entry given that iOS clients are supposed to have one to begin with.

Sorry but i've no idea to work/try with bytecode xD

oSumAtrIX:
I tried spoofing the user agent, but the playback speed menu entry was still missing. I couldn't test HDR yet, though.

zainarbani:
Can you push the test commit? I'll try

0xrxL:
There's another issue with iOS spoof: you can't listen a live stream video in background.

zainarbani:
> I tested it locally with a proxy, no code changes

Which user-agent did you use btw?

Pamilg8:
I think this PR will have to wait a bit. There seems to be a lot of bugs

src/main/kotlin/app/revanced/patches/youtube/misc/fix/playback/SpoofClientPatch.kt
```diff
@@ -168,6 +168,22 @@ object SpoofClientPatch : BytecodePatch(
...
+            val getClientModelIndex = it.indexOfFirstInstruction {
+                getReference<FieldReference>().toString() == "Landroid/os/Build;->MODEL:Ljava/lang/String;"
+            }
+
+            // The next IPUT_OBJECT instruction after getting the client model is setting the client model field.
+            instructions.subList(
+                getClientModelIndex,
+                instructions.lastIndex,
+            ).first { instruction ->
+                instruction.opcode == Opcode.IPUT_OBJECT
```
Francesco146, line 183: on newer version of YT the instruction might be `SGET_OBJECT`

0xrxL:
> Playback speed menu is also fixed if the client is spoofed to Android VR (client 28, version "1.37").
> 
> All video resolutions appear to be present (I can't tell if HDR is working or not). Audio only playback of livestreams works.
> 
> Only thing I found that doesn't work is player swipe gestures.
> 
> Not sure if playback is broken or not, as I've never had that issue.

Try this for HDR:

https://youtu.be/njX2bu-_Vw4

0xrxL:
> > Try this for HDR:
> > https://youtu.be/njX2bu-_Vw4
> 
> HDR is not present with android vr, at least not without additional changes. I did not try changing the client device name since I can't seem to find what the name should be. Edit: the model appears to be [`Quest`](https://gist.github.com/hxhb/0c72c436044e97f3f11ea44ad7f92912#file-oculus-quest-build-props-L22), but setting this did not change anything.
> 
> Just noticed an odd bug though, where if playback is paused it will randomly resume playback (including if the app is paused in the background). This doesn't show up with any other client I tried.
> 
> Some of the other iOS and Android clients also worked, but all that I tried were capped at 720p.

Have you tried with this?

Screenshot_20240523_191816_Samsung Internet Beta

0xrxL:
> > Have you tried with this?
> 
> Very mostly broken. Incognito fails. Non incognito playback starts with a weird brief popup that looks bizarrely like html 1.0 (even though that makes no sense). no high resolution video.

Ok, so it can be possible to rely only on few clients, if the point is to limiting the bugs.

...and i'm afraid that HDR can't be fixed, because inotia00 said me that codec to correctly decode this videos type is missing in the android package.

Pamilg8:
it seems that the string has not been updated. When the iOS client is opened, it is shown that there are side windows.

Edgars-Cirulis:
Screenshot_20240526_005401_YouTube
HDR missing on S23 Ultra.

Tokudas:
Can this application be fixed?
https://play.google.com/store/apps/details?id=com.commsource.beautyplus
Screenshot_2024-05-26-19-06-32-47

PR #3253, opened 2024-05-30
fix(Third party Reddit clients): Spoof user agent to workaround Reddit API issues

OctoNezd:
Reddit started blocking all requests that contain word "android" in them: https://old.reddit.com/r/redditdev/comments/1d3spoc/error_getting_submitted_with_mobile_useragent/

Apps started getting 400 or 503, if they have "android" in their User-Agent header:

image

image

Weirdly, 400 requests get "Internal Server Error":

image

While it might be possible it is temporary outage, but it happens only with "android" in user-agent, which is suspicious, at least.


BoxxiDev:
If you can extend this to Joey as well I'd be super-appreciative. Great work regardless.

src/main/kotlin/app/revanced/patches/reddit/customclients/boostforreddit/api/SpoofClientPatch.kt
```diff
@@ -24,4 +24,15 @@ object SpoofClientPatch : BaseSpoofClientPatch(
             """
         )
     }
+
+    override fun Set<MethodFingerprintResult>.patchUserAgent(context: BytecodeContext) {
+        // Use a random user agent.
+        val randomName = (0..100000).random()
```
oSumAtrIX, line 30: Why is a random integer used here as a name but in the other patch the full user agent string?

OctoNezd: Boost uses [JRAWs UserAgent](https://github.com/mattbdean/JRAW/blob/fa1efa33729d42c6d2fc52f4cf67eabcb75e3be9/lib/src/main/kotlin/net/dean/jraw/http/UserAgent.kt#L25) class that takes in platform, app name, app version and developers username, and at time I was making fix it looked more proper to me to swap platform instead of changing whole User-Agent.

Sync uses hardcoded user-agent, instead of generating it on fly.

oSumAtrIX (approved):

oSumAtrIX:
Sorry the approval was by mistake

oSumAtrIX (commented):
This PR was made because Reddit broke their oAuth API for the Android platform. In the past Reddit has caused the same issue and every time it resolved itself after a while. Sometimes a desktop browser user agent was required to bypass issues with logging in for example. For that reason this PR might be obsolete soon, if Reddit fixes their issue. Until then, Reddit should be contacted first and asked about the issue.

OctoNezd:
Perhaps make randomizing user-agent an option? In case reddit fixes it but then it might break it again which would require this PR again.

oSumAtrIX:
That would be overengineering around a problem ReVanced isn't supposed to look at, which is why I'm even hesitant with this PR. Reddit can make any unpredictable change, not just to the user agent.

OctoNezd:
> If you can extend this to Joey as well I'd be super-appreciative. Since Joey is based on Boost as far as I understand it I hope it is relatively similar. Great work regardless.

Implemented in https://github.com/ReVanced/revanced-patches/pull/3253/commits/e1754f3d54c183c1b11dcf416cd1f7e389959d37

oSumAtrIX:
RiF seems to be partially affected (https://github.com/ReVanced/revanced-patches/issues/3256), although I haven't checked, if it is related to the user agent and can be fixed with this PR

thryyy:
I'm getting "Invalid Request to Oauth API" even after patching with the updated patch from this PR

Chezpuf:
I am having the same issues with Baconreader. Would this fix work there too?

src/main/kotlin/app/revanced/patches/reddit/customclients/joeyforreddit/api/fingerprints/AuthUtilityUserAgent.kt
```diff
@@ -0,0 +1,11 @@
+package app.revanced.patches.reddit.customclients.joeyforreddit.api.fingerprints
+
+import app.revanced.patcher.fingerprint.MethodFingerprint
+
+object AuthUtilityUserAgent : MethodFingerprint (
+    strings = setOf("2.1.6.5"),
```
oSumAtrIX, line 6: Is this the app version string? If so, if the app updates this string would change. Probably should not include in the fingerprint

ianroberts-bizclik:
> RiF seems to be partially affected (#3256), although I haven't checked, if it is related to the user agent and can be fixed with this PR

RIF needs line 55 changing from:

`val userAgent = "android:app.revanced.$randomName:v1.0.0 (by /u/revanced)"`

To:

`val userAgent = "$randomName:app.revanced.$randomName:v1.0.0 (by /u/revanced)"`

oSumAtrIX:
RiF is not included in this PR, not sure what file you are referencing.

ianroberts-bizclik:
> RiF is not included in this PR, not sure what file you are referencing.

https://github.com/ReVanced/revanced-patches/blob/21f87db6c5ee2467979edbf352f704a9fc72efca/src/main/kotlin/app/revanced/patches/reddit/customclients/redditisfun/api/SpoofClientPatch.kt#L55

Unsure why it wouldn't be included as the useragent from your previous patch includes "android"

src/main/kotlin/app/revanced/patches/reddit/customclients/joeyforreddit/api/fingerprints/AuthUtilityUserAgent.kt
BoxxiDev: These Reddit applications being patched are all abandoned by the developers (due to changes by Reddit.) So 2.1.6.5 is the last version of Joey before it was abandoned. Maybe it is possible to simply make this patch require this specific version? Or, even better - I'm sure it's possible to fingerprint this method in some other way; maybe you can point him to the best possible fingerprint option based on the below? The last method returns the user agent and needs to be overridden.

image

BoxxiDev: Looks like this in smali:
image

oSumAtrIX: The opcode pattern can be fingerprinted

BoxxiDev:
Thank you for the Joey fix! I think maybe an elegant solution to this would be to let the user specify their own user-agent instead of accepting a random one. That would make it more future-proof as well. But I think there's a lot of people that would love it as-is simply to get their Reddit apps working on mobile working again.

Jo0:
> I think maybe an elegant solution to this would be to let the user specify their own user-agent instead of accepting a random one (just like client ID is currently a manual input). 

This definitely feels like the play.

oSumAtrIX:
The way it is currently done is good enough. With patching your app, you compile a "new" application in context of Reddit oAuth applications which is why it receives your personal unique user agent.

oSumAtrIX (approved):

oSumAtrIX:
Because a lot of reports are coming in. It's easier to merge this and then revert if necessary, than keep it broken and wait until its fixed by Reddit.

zimmertr:
FYI, RIF just started working for me again.

oSumAtrIX:
[As expected](https://github.com/ReVanced/revanced-patches/pull/3253#pullrequestreview-2089382033). Spoofing the user agent won't break it, but it's unnecessary if it doesn't fix anything.

CodingDinosaur:
RIF is still not working for me, so I'm not sure if this has been adjusted by Reddit or not.  I will update if it starts working.

zimmertr:
You're correct, I was wrong. Comments still aren't loading. 

I believe that when things originally broke, the app would not load a feed at all. When it started to do so, I thought the issue was resolved.

TortoiseFather:
Hi all,
Sorry for noob question but will I need to make a new Reddit app to patch rif?

I'm just following [https://www.reddit.com/r/Android/comments/14niwgu/revanced_patches_for_boost_infinity_rif_is_fun/](this guide) but I imagine its out of date

ianroberts-bizclik:
Forked the repo and forced a release live to test, working as expected for RIF with the new patch

ksnaden:
Thanks, I used revanced-cli and patched RiF. Works.  Much appreciated.

AElimrani:
Awesome!! Can confirm this works, just have to use the same patch as before. Maybe adding a new name might be helpful (instead of just spoof client Id). Thanks a lot!

kokroo:
I'm confused. Could someone please explain how to apply this newest patch to RIF?

CodingDinosaur:
The new patch hasn't been merged to main yet.  #3234

You won't be able to use the patch until it's merged and released.  (Unless you use an advanced method such as the cli or a fork, but if you're not sure how to do this just wait for the release)

e-xception:
Any chance we could also get Slide for Reddit updated? Many thanks!

Duramoin:
@ksnaden hello, could you please post your whole command for rif? thanks.. I am getting error.. i use jars and integration with latest version and rif apk from apkmirror

`java -jar cli.jar patch rif.apk -b patches.jar -m integrations.apk -o rif1.apk`

`INFO: Aligning APK
org.bouncycastle.jcajce.provider.keystore.bc.BcKeyStoreSpi$BCKeyStoreException: java.io.IOException: Error initialising store of key store: java.lang.SecurityException: JCE cannot authenticate the provider BC
        at org.bouncycastle.jcajce.provider.keystore.bc.BcKeyStoreSpi.engineSetKeyEntry(Unknown Source)
        at java.base/java.security.KeyStore.setKeyEntry(KeyStore.java:1191)
        at app.revanced.library.ApkSigner.newKeyStore(ApkSigner.kt:56)
        at app.revanced.library.ApkUtils.newPrivateKeyCertificatePair(ApkUtils.kt:114)
        at app.revanced.library.ApkUtils.signApk(ApkUtils.kt:167)
        at app.revanced.cli.command.PatchCommand.run(PatchCommand.kt:321)
        at picocli.CommandLine.executeUserObject(CommandLine.java:2026)
        at picocli.CommandLine.access$1500(CommandLine.java:148)
        at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2461)
        at picocli.CommandLine$RunLast.handle(CommandLine.java:2453)
        at picocli.CommandLine$RunLast.handle(CommandLine.java:2415)
        at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2273)
        at picocli.CommandLine$RunLast.execute(CommandLine.java:2417)
        at picocli.CommandLine.execute(CommandLine.java:2170)
        at app.revanced.cli.command.MainCommandKt.main(MainCommand.kt:12)
Caused by: java.io.IOException: Error initialising store of key store: java.lang.SecurityException: JCE cannot authenticate the provider BC
        at org.bouncycastle.jcajce.provider.keystore.bc.BcKeyStoreSpi.makePBECipher(Unknown Source)
        at org.bouncycastle.jcajce.provider.keystore.bc.BcKeyStoreSpi$StoreEntry.<init>(Unknown Source)
        ... 15 more`

oSumAtrIX:
For For support, move to one of our links at https://revanced.app.

PR #3259, opened 2024-05-31
fix(YouTube/Spoof client): player gestures not working when spoofing with `Android VR` client

kazimmt:
Courtesy: https://github.com/inotia00/revanced-patches/commit/0b35694d18958f3787128bb88187db661c098dd4


src/main/kotlin/app/revanced/patches/youtube/misc/fix/playback/SpoofClientPatch.kt
```diff
@@ -115,6 +120,28 @@ object SpoofClientPatch : BytecodePatch(
 
         // endregion
 
+        // region fix player gesture.
+
+        PlayerGestureConfigSyntheticFingerprint.resultOrThrow().let {
+            arrayOf(3, 9).forEach { offSet ->
+                it.mutableMethod.apply {
```
inotia00, line 127: To access the method to hook, you should use MethodWalker
Also offSet is not being used

It should be changed as follows:
```
val endIndex = it.scanResult.patternScanResult!!.endIndex
val targetMethod = context.toMethodWalker(it.method)
                .nextMethod(endIndex - offSet, true)
                .getMethod() as MutableMethod
targetMethod.apply {
...
}
```

src/main/kotlin/app/revanced/patches/youtube/misc/fix/playback/fingerprints/PlayerGestureConfigSyntheticFingerprint.kt
```diff
@@ -0,0 +1,53 @@
...
+import app.revanced.patches.youtube.misc.fix.playback.fingerprints.PlayerGestureConfigSyntheticFingerprint.indexOfDownAndOutAllowedInstruction
+import app.revanced.util.getReference
+import app.revanced.util.indexOfFirstInstruction
+import com.android.tools.smali.dexlib2.AccessFlags
+import com.android.tools.smali.dexlib2.Opcode
+import com.android.tools.smali.dexlib2.iface.Method
+import com.android.tools.smali.dexlib2.iface.reference.MethodReference
+
+
+@FuzzyPatternScanMethod(5)
```
inotia00, line 15: You can remove `FuzzyPatternScanMethod`
(Compatibility of the `Spoof client` patch is from YouTube v18.37.36 to YouTube v19.11.43)

src/main/kotlin/app/revanced/patches/youtube/misc/fix/playback/SpoofClientPatch.kt
```diff
@@ -115,6 +121,33 @@ object SpoofClientPatch : BytecodePatch(
...
+
+                        val index = implementation!!.instructions.lastIndex
+                        val register = getInstruction<OneRegisterInstruction>(index).registerA
+
+                        addInstructions(
+                            index,
+                            """
+                            invoke-static {v$register}, $INTEGRATIONS_CLASS_DESCRIPTOR->enablePlayerGesture(Z)Z
+                            move-result v$register
+                        """
```
oSumAtrIX, line 143: This needs to be formatted

src/main/kotlin/app/revanced/patches/youtube/misc/fix/playback/SpoofClientPatch.kt
```diff
@@ -115,6 +121,33 @@ object SpoofClientPatch : BytecodePatch(
 
         // endregion
 
+        // region fix player gesture.
+
+        PlayerGestureConfigSyntheticFingerprint.resultOrThrow().let {
+            val endIndex = it.scanResult.patternScanResult!!.endIndex
```
oSumAtrIX, line 127: I was hoping to refactor this before merging. It's quite convoluted and could've been replaced with something simpler probably

oSumAtrIX: Oh yeah, in that case it'll be necessary to get to it via another method. I'll try my shot in a bit

Marekk2k:
I would like to ask when the option to set the video speed in YouTube ReVanced will be restored? I miss this option so much :(

src/main/kotlin/app/revanced/patches/youtube/misc/fix/playback/SpoofClientPatch.kt
inotia00: I also wanted to use a more ideal fingerprint to identify the Method to which the patch would be applied
Because the current fingerprint is overly complex, I thought it might break easily when the YouTube version is upstreamed

So I also tried refactoring, but there was no success
As @LisoUseInAIKyrios explained, there was no way to identify which Method the patch would be applied to without using MethodWalker

Instead, I tested all major versions from YouTube v18.29.38 to YouTube v19.20.34 using the current fingerprint, and confirmed that the fingerprint does not break easily

If I can find a more ideal fingerprint, I will mention it here then

PR #3312, opened 2024-06-07
fix(YouTube - Bottom buttons): Adjust bottom buttons layout

MarcaDian:
Fixes the clicking effect of the bottom buttons by adding the layout_marginBottom attribute (I think 6.0dip will be fine, see screenshots). 
Resizes the button's container to fit better relative to each other to somewhat prevent the stacking effect on other elements (60x60 originally, 48x48 now).
It should also be noted that in YouTube 19.09+ fullscreen_button has been moved from youtube_controls_bottom_ui_container to other layouts, so it is necessary to change the attributes for them.
<details>
 <summary>Before</summary>

IMG_20240607_132930
</details>
<details>
 <summary>After</summary>

IMG_20240607_132949

</details>


oSumAtrIX:
Why does the incorrect margin happen to begin with?

oSumAtrIX:
In that case, this should be first reported to YouTube and fixed by them. The patch is overly complex for a minuscule gain. Thanks regardless.

PR #3371, opened 2024-06-24
feat(Instagram): Add `Hide profile ads` patch

tymmesyde:
Add a patch that removes ads on Instagram profiles

### DIFF
--- a/src/main/kotlin/app/revanced/patches/instagram/patches/ads/profile/HideProfileAdsPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/instagram/patches/ads/profile/HideProfileAdsPatch.kt
@@ -25,17 +25,16 @@ object HideProfileAdsPatch : BytecodePatch(
     override fun execute(context: BytecodeContext) {
         ProfileAdInjectorFingerprint.result?.let {
             it.mutableMethod.apply {
-                val firstInstructionIndex = it.scanResult.patternScanResult!!.startIndex
-                val conditionInstructionIndex = it.scanResult.patternScanResult!!.endIndex
+                val conditionalJumpIndex = it.scanResult.patternScanResult!!.endIndex
 
-                val conditionInstructionLabel = getInstruction<BuilderInstruction21t>(conditionInstructionIndex).target
+                val conditionalJumpLabel = getInstruction<BuilderInstruction21t>(conditionalJumpIndex).target
 
-                // Replace the first instruction of the method with a goto to the label from the condition instruction
+                // Replace this conditional jump by a goto
                 replaceInstruction(
-                    firstInstructionIndex,
+                    conditionalJumpIndex,
                     BuilderInstruction10t(
                         Opcode.GOTO,
-                        conditionInstructionLabel,
+                        conditionalJumpLabel,
                     ),
                 )
             }
--- a/src/main/kotlin/app/revanced/patches/instagram/patches/ads/profile/HideProfileAdsPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/instagram/patches/ads/profile/HideProfileAdsPatch.kt
@@ -6,7 +6,7 @@ import app.revanced.patcher.extensions.InstructionExtensions.replaceInstruction
 import app.revanced.patcher.patch.BytecodePatch
 import app.revanced.patcher.patch.annotation.CompatiblePackage
 import app.revanced.patcher.patch.annotation.Patch
-import app.revanced.patches.instagram.patches.ads.profile.fingerprints.HideProfileAdsFingerprint
+import app.revanced.patches.instagram.patches.ads.profile.fingerprints.ProfileAdInjectorFingerprint
 import app.revanced.util.exception
 import com.android.tools.smali.dexlib2.Opcode
 import com.android.tools.smali.dexlib2.builder.instruction.BuilderInstruction10t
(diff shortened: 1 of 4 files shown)


src/main/kotlin/app/revanced/patches/instagram/patches/ads/profile/fingerprints/HideProfileAdsFingerprint.kt
```diff
@@ -0,0 +1,17 @@
+package app.revanced.patches.instagram.patches.ads.profile.fingerprints
+
+import app.revanced.patcher.extensions.or
+import app.revanced.patcher.fingerprint.MethodFingerprint
+import com.android.tools.smali.dexlib2.AccessFlags
+import com.android.tools.smali.dexlib2.Opcode
+
+internal object HideProfileAdsFingerprint : MethodFingerprint(
```
oSumAtrIX, line 8: Please rename this fingerprint to the method it resolves to. If the method name is obfuscated, think of what the method could've been called, for example, "ShowProfileAdFingerprint" (Use a better name if possible).

src/main/kotlin/app/revanced/patches/instagram/patches/ads/profile/HideProfileAdsPatch.kt
```diff
@@ -0,0 +1,44 @@
...
+@Suppress("unused")
+object HideProfileAdsPatch : BytecodePatch(
+    setOf(
+        HideProfileAdsFingerprint,
+    ),
+) {
+    override fun execute(context: BytecodeContext) {
+        HideProfileAdsFingerprint.result?.let {
+            it.mutableMethod.apply {
+                val firstInstructionIndex = it.scanResult.patternScanResult!!.startIndex
```
oSumAtrIX, line 28: You can simply replace the instruction at `endIndex`, then you don't need this line of code.

src/main/kotlin/app/revanced/patches/instagram/patches/ads/profile/HideProfileAdsPatch.kt
```diff
@@ -0,0 +1,44 @@
...
+import app.revanced.patcher.patch.annotation.CompatiblePackage
+import app.revanced.patcher.patch.annotation.Patch
+import app.revanced.patches.instagram.patches.ads.profile.fingerprints.HideProfileAdsFingerprint
+import app.revanced.util.exception
+import com.android.tools.smali.dexlib2.Opcode
+import com.android.tools.smali.dexlib2.builder.instruction.BuilderInstruction10t
+import com.android.tools.smali.dexlib2.builder.instruction.BuilderInstruction21t
+
+@Patch(
+    name = "Hide profile ads",
```
oSumAtrIX, line 16: What are "profile ads"?

tymmesyde: Not sure how to call it but when you go on an instagram profile and open a post, you get a feed of all the posts and as you scroll, there are ads that are injected

oSumAtrIX: We should probably integrate this patch into the existing patch and call the patch "Feed" patch, given that this is a feed as well. The description can then mention that the patch hides ads in the timeline and profile feed.

src/main/kotlin/app/revanced/patches/instagram/patches/ads/profile/HideProfileAdsPatch.kt
tymmesyde: Right, it's just that it is invoking something just before the conditional jump and thought it was best to skip it

oSumAtrIX: It could have a negative effect. Do you know what that method does?

src/main/kotlin/app/revanced/patches/instagram/patches/ads/profile/HideProfileAdsPatch.kt
tymmesyde: Yeah why not but it seems that the other patch is not up to date anymore

src/main/kotlin/app/revanced/patches/instagram/patches/ads/profile/HideProfileAdsPatch.kt
tymmesyde: No, i did what you suggested

src/main/kotlin/app/revanced/patches/instagram/patches/ads/profile/HideProfileAdsPatch.kt
oSumAtrIX: That's due to the patch not working on the latest version

src/main/kotlin/app/revanced/patches/instagram/patches/ads/profile/HideProfileAdsPatch.kt
oSumAtrIX: I suggested it because I didn't know there was a method being called. Should probably be looked into, given you previously had decided to not call it

tymmesyde:
Closed in favor of https://github.com/ReVanced/revanced-patches/pull/3380

PR #3379, opened 2024-06-24
fix(Boost for reddit - Fix missing audio in video downloads): Fix invalid string replacement

OctoNezd:
Some refactor broke the audio download feature, resulting in `v.redd.it` and `/` strings getting replaced instead of audio endpoints resulting in client not even trying to download audio. This fixes that, by removing these two strings and keeping only `/DASH_audio.mp4` and `/audio`. Both strings are unique in final version of Boost.


src/main/kotlin/app/revanced/patches/reddit/customclients/boostforreddit/fix/downloads/FixAudioMissingInDownloadsPatch.kt
```diff
@@ -25,7 +25,7 @@ object FixAudioMissingInDownloadsPatch : BytecodePatch(
     )
     override fun execute(context: BytecodeContext) {
         DownloadAudioFingerprint.resultOrThrow().let { result ->
-            result.scanResult.stringsScanResult!!.matches.take(2).forEach { match ->
+            result.scanResult.stringsScanResult!!.matches.forEach { match ->
```
oSumAtrIX, line 28: I think a loop isn't necessary, just take the first string

OctoNezd: We need to replace both strings. Some videos dont have DASH_AUDIO_128, only DASH_AUDIO_64.

Eagle1337:
I tested it using the 4.11.0-dev2 patches which afaik should have the fix and it still doesn't work for me

OctoNezd:
Tested patch with `java -jar .\revanced-cli-4.6.0-all.jar patch -b revanced-patches-4.11.0-dev.2.jar -m revanced-integrations-1.11.0.apk --options=C:\Users\octo\rev\options.json C:\Users\octo\rev\src\boost.apk -d`, audio is downloading correctly. Can you share link to post where audio is not downloading correctly?

Eagle1337:
> Tested patch with `java -jar .\revanced-cli-4.6.0-all.jar patch -b revanced-patches-4.11.0-dev.2.jar -m revanced-integrations-1.11.0.apk --options=C:\Users\octo\rev\options.json C:\Users\octo\rev\src\boost.apk -d`, audio is downloading correctly. Can you share link to post where audio is not downloading correctly?
[[bb](https://www.reddit.com/r/holdmycatnip/comments/1dtovog/cats_can_sleep_anywhere_they_want/)](https://www.reddit.com/r/holdmycatnip/comments/1dtovog/cats_can_sleep_anywhere_they_want/)

I'm not great with  cli, so I used rvx (after telling it to use revanced patches) and told it to use the pre-release patches. I'll probably just wait until the patches get pushed to main and retry then.

Eagle1337:
It seems to have broken again, I'm not getting audio when saving videos from reddit again.

OctoNezd:
I will look into fixing it on holidays. Need to replace it with an *actual* fix - parsing the MPD and getting audio from it instead of relying on hardcoded paths.

Issue #1221, opened 2023-03-26
feat(windy): windy.com app premium patch 


SodaWithoutSparkles:
### Application

[windy.com app](https://play.google.com/store/apps/details?id=com.windyty.android), aka "red windy"

### Issue

Some features that was free is now behind the premium paywall. For example, 11 day forcast, 12h weather radar and satellite imagery, etc.

* this app is not the same as the windy.app app requested before. The windy.app app is also known as "blue windy".

### Patch

Unlock premium or unlock the features that were available before.

### Motivation

Windy.com app is a weather forcasting app that is used by many people. It features a smooth UI 

### Acknowledgements

- [X] I have searched the existing issues and this is a new and no duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this issue properly.

PR #3397, opened 2024-06-30
fix(Windy - Unlock pro): Use correct package name

Shoggomo:


oSumAtrIX (approved):

Gopnk:
Why was the package name changed?
com.windyty is for a different app "red windy" called windy.com which has a premium plan.
co.windyapp is for the app "blue windy" windy.app which has a pro plan for which this patch worked.

https://github.com/ReVanced/revanced-patches/issues/1221

Shoggomo:
Oh damn, I didn't realize this was was for another app. Sorry, then this change should be reversed of course.

PR #3386, opened 2024-06-26
feat(Soundcloud): Add hide ads and disable telemetry patch

patchink0:
Removes all ads and disables Soundcloud's in-house telemetry system.
This patch does not give access to any premium features.

PR #3407, opened 2024-07-02
feat(SoundCloud): Add `Enable offline sync` patch

BenCat07:
Depends on #3386

Enables the offline download Feature, allowing listening to tracks without internet

It patches the request Involved to use the https stream endpoint instead. (from `/tracks/(id)/offline_sync` -> `/tracks/(id)/streams/https`)

There 3 (purely cosmetic) headers missing on this endpoint that crash SoundCloud, they get replaced with empty Strings if detected as null.


src/main/kotlin/app/revanced/patches/soundcloud/offline/AllowOfflineSyncPatch.kt
```diff
@@ -0,0 +1,100 @@
...
+            /**
+             * HTTPS Stream has one major flaw. It does not return 3 specific headers, which crashes Soundcloud.
+             *
+             * Since those are only cosmetic though, we'll simply move "" into the local Variables when appropiate.
+             */
+            // These indices Represent the opcodes on which "if-eqz" is performed. Before they run,
+            // If we *do* detect the null and instead move "" into there.
+            val opcodeSize = 2
+            val firstZeroCheck = 4
+            // Adding instructions shifts our indices around, adjust appropiately
```
oSumAtrIX, line 69: Add them in reverse from last to first. This way no shifting occurs.

src/main/kotlin/app/revanced/patches/soundcloud/offline/AllowOfflineSyncPatch.kt
```diff
@@ -0,0 +1,100 @@
...
+            val secondZeroCheck = 8 + opcodeSize
+            val thirdZeroCheck = 12 + 2 * opcodeSize
+            addInstructionsWithLabels(
+                firstZeroCheck,
+                """
+                    if-nez v0, :skip1
+                    const-string v0, ""
+            """,
+                ExternalLabel("skip1", getInstruction(firstZeroCheck)),
```
oSumAtrIX, line 78: As the headers are cosmetic, as described in the PR, emptying the strings always is simpler without a null check.

src/main/kotlin/app/revanced/patches/soundcloud/ad/fingerprints/InterceptFingerprint.kt
```diff
@@ -0,0 +1,22 @@
+package app.revanced.patches.soundcloud.ad.fingerprints
+
+import app.revanced.patcher.fingerprint.MethodFingerprint
+import com.android.tools.smali.dexlib2.AccessFlags
+import com.android.tools.smali.dexlib2.Opcode
+
+internal object InterceptFingerprint : MethodFingerprint(
```
oSumAtrIX, line 7: Please name this fingerprint according to the method it resolves to. "Intercept" is not really descriptive.

src/main/kotlin/app/revanced/patches/soundcloud/offline/AllowOfflineSyncPatch.kt
BenCat07: You're right, that's the correct way to do it

src/main/kotlin/app/revanced/patches/soundcloud/offline/AllowOfflineSyncPatch.kt
BenCat07: fair enough

src/main/kotlin/app/revanced/patches/soundcloud/ad/fingerprints/InterceptFingerprint.kt
BenCat07: Sorry this is based on the commits of the other soundcloud PR, i will rebase onto it occasionally, but this is not my thing to fix

oSumAtrIX: Oh I see, I was wondering how that patch came into this PR

BenCat07: That's why this is marked as a draft aswell, and "Depends on ...", if i had the option like in gitlab it would not show these changes, and only come available after the dependency is merged

src/main/kotlin/app/revanced/patches/soundcloud/offline/AllowOfflineSyncPatch.kt
BenCat07: It might also be worth a thought to have just 2 add instruction blocks, one that just moves "" into all the strings, and another that just delets the if-eqz checks... but i think this is more readable:
```java
            val firstZeroCheck = 4
            val secondZeroCheck = 8
            val thirdZeroCheck = 12
            addInstruction(thirdZeroCheck, "const-string v6, \"\"")
            addInstruction(secondZeroCheck, "const-string v2, \"\"")
            addInstruction(firstZeroCheck, "const-string v0, \"\"")```

Comment (oSumAtrIX) [reply src/main/kotlin/app/revanced/patches/soundcloud/offline/AllowOfflineSyncPatch.kt:78]: Can you just remove the if instructions? If they jump over the wanted call this should work

Comment (BenCat07) [reply src/main/kotlin/app/revanced/patches/soundcloud/offline/AllowOfflineSyncPatch.kt]: > Can you just remove the if instructions? If they jump over the wanted call this should work

Nope, it checks for null again inside the constructor of the object those are passed to, and again every time said object is copied. So in end effect i would instead have to remove 9 instructions over 2 different files and 3 different methods

Comment (BenCat07) [reply src/main/kotlin/app/revanced/patches/soundcloud/offline/AllowOfflineSyncPatch.kt]: The object is still used, just not for these three things. I could however jump over the ifs and load "" into the variables passed in that would work.

Comment (oSumAtrIX) [reply src/main/kotlin/app/revanced/patches/soundcloud/offline/AllowOfflineSyncPatch.kt]: In that case the current approach is fine

Comment (BenCat07) [reply src/main/kotlin/app/revanced/patches/soundcloud/offline/AllowOfflineSyncPatch.kt]: > > Can you just remove the if instructions? If they jump over the wanted call this should work
> 
> Nope, it checks for null again inside the constructor of the object those are passed to, and again every time said object is copied. So in end effect i would instead have to remove 9 instructions over 2 different files and 3 different methods

In fact more than that since it does not use a simple if-eqz there but a static method call with extra parameters.
```java
        Intrinsics.checkNotNullParameter(urn, "urn");
        Intrinsics.checkNotNullParameter(preset, "preset");
        Intrinsics.checkNotNullParameter(quality, "quality");
        Intrinsics.checkNotNullParameter(mimeType, "mimeType");
```

oSumAtrIX (approved):

moteb757:
F

PR #3418, opened 2024-07-06
feat(GPS Status): unlock PRO and disable ads

programminghoch10:
my first time contributing to revanced
my first time smali
my first time kotlin
my first time frida
my first time jadx
only took about 10h to complete

please tell me if anything can be improved

### DIFF
--- a/src/main/kotlin/app/revanced/patches/com/eclipsim/gpsstatus2/UnlockProPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/com/eclipsim/gpsstatus2/UnlockProPatch.kt
@@ -1,0 +1,53 @@
+package app.revanced.patches.com.eclipsim.gpsstatus2
+
+import app.revanced.patcher.data.BytecodeContext
+import app.revanced.patcher.extensions.InstructionExtensions.addInstructions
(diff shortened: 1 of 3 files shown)


oSumAtrIX:
This is unfortunate, but our contribution guidelines permit pro unlock patches :/

https://github.com/ReVanced/revanced-patches/blob/main/CONTRIBUTING.md#%EF%B8%8F-guidelines-for-requesting-or-contributing-patches

programminghoch10:
Well in this case the ads are disabled because pro is activated.
I could develop a separate ads hiding patch, but since the ads aren't that obtrusive I don't really want to bother.

oSumAtrIX:
> Well in this case the ads are disabled because pro is activated.

I noticed. You may be better off with publishing the patch yourself. We have this policy because it caused trouble for us in the past. WIth the new version of ReVanced Manager, just like in CLI you will be able to use multiple patches files at a time so people can use anyone patches side by side: https://github.com/ReVanced/revanced-manager/milestone/1 

Thanks regardless for the PR!

PS: If you want to accelerate the development of the new version of ReVanced Manager, we would welcome any contributions there too!

PR #3420, opened 2024-07-08
feat(Duolingo): Add `Unlock Duolingo Super` patch back again

notjosh:
Hiya! :wave:

A little while ago the Duolingo patch was removed after it stopped working. In this PR I've fixed the old code, and made it work with newer versions of Duolingo.

It's still a little fragile (tho I guess all of ReVanced patches are inherently fragile!), but seems to be working reliably enough for now.

### DIFF
--- a/api/revanced-patches.api
+++ b/api/revanced-patches.api
@@ -193,6 +193,12 @@ public final class app/revanced/patches/cieid/restrictions/root/BypassRootChecks
 	public synthetic fun execute (Lapp/revanced/patcher/data/Context;)V
 }
 
+public final class app/revanced/patches/duolingo/unlocksuper/UnlockDuolingoSuperPatch : app/revanced/patcher/patch/BytecodePatch {
+	public static final field INSTANCE Lapp/revanced/patches/duolingo/unlocksuper/UnlockDuolingoSuperPatch;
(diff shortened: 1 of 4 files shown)


oSumAtrIX:
This is unfortunate because our contribution guidelines do not permit pro-unlock patches :/

https://github.com/ReVanced/revanced-patches/blob/main/CONTRIBUTING.md#%EF%B8%8F-guidelines-for-requesting-or-contributing-patches

But this does not mean that you need to scrap your patch, because you can publish your own patches: https://github.com/ReVanced/revanced-patches-template (checkout dev)

I would personally make an exception for this patch according to

> We reserve the right to make exceptions for patches that provide significant value.

but I'd like to see if @LisoUseInAIKyrios also approves

oSumAtrIX:
Thanks regardless

notjosh:
Thanks for the feedback! I read the guidelines before submitting, and ultimately my goal was to achieve this:

> Ad-Blocking: Patches aimed at enhancing user privacy and blocking intrusive advertisements are appreciated

I don't really care about the actual "super" features, just to block the ads (which this PR does). Would it be worth rewriting the patch specifically to skip the ad dialogs, if there's a hook I can find to that?

oSumAtrIX:
That makes sense. If you can submit an af blocking patch, I'd be happy to merge it!

PR #3422, opened 2024-07-08
feat(Duolingo): Add patches for skipping ads and accessing debug menu

notjosh:
Following up from #3420, I found a reasonably reliable way to skip ads within the app without impacting any of the other features/limitations of free users based on some debug flags within the app.

While I was there, I noticed a way to unlock the debug menu in production builds. I've included that in this PR, but set the `used = false` so it's not available by default, as I don't think generally people should be touching those settings. But it might be useful for some folks looking to target certain language lessons, so I think it's worth keeping it as an option.

### DIFF
--- a/api/revanced-patches.api
+++ b/api/revanced-patches.api
@@ -193,6 +193,12 @@ public final class app/revanced/patches/cieid/restrictions/root/BypassRootChecks
 	public synthetic fun execute (Lapp/revanced/patcher/data/Context;)V
 }
 
+public final class app/revanced/patches/duolingo/ad/RemoveDuolingoAdsPatch : app/revanced/patcher/patch/BytecodePatch {
+	public static final field INSTANCE Lapp/revanced/patches/duolingo/ad/RemoveDuolingoAdsPatch;
+	public fun execute (Lapp/revanced/patcher/data/BytecodeContext;)V
+	public synthetic fun execute (Lapp/revanced/patcher/data/Context;)V
+}
+
 public final class app/revanced/patches/facebook/ads/story/HideStoryAdsPatch : app/revanced/patcher/patch/BytecodePatch {
 	public static final field INSTANCE Lapp/revanced/patches/facebook/ads/story/HideStoryAdsPatch;
 	public fun execute (Lapp/revanced/patcher/data/BytecodeContext;)V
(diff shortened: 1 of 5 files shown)


oSumAtrIX (commented):
Looking good. I'll push a refactor/style commit soon and should be ready then, thanks!

src/main/kotlin/app/revanced/patches/duolingo/debug/EnableDebugMenuPatch.kt
```diff
@@ -0,0 +1,36 @@
...
+)
+@Suppress("unused")
+object EnableDebugMenuPatch : BytecodePatch(
+    setOf(InitializeBuildConfigProviderFingerprint)
+) {
+    override fun execute(context: BytecodeContext) {
+        InitializeBuildConfigProviderFingerprint.resultOrThrow().mutableMethod.apply {
+            val setIsDebugBuildIndex = getInstructions().firstOrNull {
+                it.opcode == Opcode.IPUT_BOOLEAN
+            } as? BuilderInstruction22c ?: throw InitializeBuildConfigProviderFingerprint.exception
```
oSumAtrIX, line 28: This can be replaced by adding the opcode iput boolean to the fingerprint opcode pattern and then use its index

src/main/kotlin/app/revanced/patches/duolingo/ad/DisableAdsPatch.kt
```diff
@@ -0,0 +1,42 @@
...
+        //
+        // MonetizationDebugSettings has a boolean value for "disableAds".
+        // OnboardingState has a getter to check if the user has any "adFreeSessions".
+        // SharedPreferences has a debug boolean value with key "disable_ads", which maps to "DebugCategory.DISABLE_ADS".
+        //
+        // MonetizationDebugSettings seems to be the most general setting to work fine.
+        InitializeMonetizationDebugSettingsFingerprint.resultOrThrow().mutableMethod.apply {
+            val setDisableAdsIndex = getInstructions().firstOrNull {
+                it.opcode == Opcode.IPUT_BOOLEAN
+            } as? BuilderInstruction22c ?: throw InitializeMonetizationDebugSettingsFingerprint.exception
```
oSumAtrIX, line 34: This can be replaced by adding the opcode iput boolean to the fingerprint opcode pattern and then use its index

oSumAtrIX (changes requested):
Please check if my refactor broke anything on your end as well

oSumAtrIX:
Is this PR still active?

oSumAtrIX (approved):

PR #3423, opened 2024-07-09
feat(GPS Status): Add `Disable ads` patch

programminghoch10:
my second time contributing to revanced
my second time smali
my second time kotlin
my second time jadx
only took about 2h to complete

please tell me if anything can be improved

after #3418 i made a new patch which only removes the ad banner

### DIFF
--- a/src/main/kotlin/app/revanced/patches/gpsstatus2/ads/DisableAdsPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/gpsstatus2/ads/DisableAdsPatch.kt
@@ -1,0 +1,27 @@
+package app.revanced.patches.gpsstatus2.ads
+
+import app.revanced.patcher.data.BytecodeContext
+import app.revanced.patcher.extensions.InstructionExtensions.removeInstructions
+import app.revanced.patcher.patch.BytecodePatch
+import app.revanced.patcher.patch.annotation.CompatiblePackage
+import app.revanced.patcher.patch.annotation.Patch
+import app.revanced.patches.gpsstatus2.ads.fingerprints.PlaceAdFingerprint
+import app.revanced.util.exception
+
+@Patch(
+    name = "Disable Ads",
+    compatiblePackages = [CompatiblePackage("com.eclipsim.gpsstatus2")]
+)
+@Suppress("unused")
+object DisableAdsPatch : BytecodePatch(
+    setOf(PlaceAdFingerprint)
+) {
+    override fun execute(context: BytecodeContext) {
+        PlaceAdFingerprint.result?.let { result ->
+            result.mutableMethod.removeInstructions(
+                result.scanResult.patternScanResult!!.endIndex,
+                1
+            )
+        } ?: throw PlaceAdFingerprint.exception
+    }
+}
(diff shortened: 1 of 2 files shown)


src/main/kotlin/app/revanced/patches/gpsstatus2/ads/fingerprints/PlaceAdFingerprint.kt
```diff
@@ -0,0 +1,37 @@
...
+        Opcode.CONST,
+        Opcode.INVOKE_VIRTUAL,
+        Opcode.MOVE_RESULT_OBJECT,
+        Opcode.CHECK_CAST,
+        Opcode.IGET_OBJECT,
+        Opcode.CONST_4,
+        Opcode.INVOKE_VIRTUAL,
+    ),
+    returnType = "V",
+    strings = listOf("banner"),
```
oSumAtrIX, line 36: Is this string present in two methods? The opcode pattern being this long may not he necessary, you could trim it down to probably the last 5-10 instructions. Adding access flags and parameters could also help

src/main/kotlin/app/revanced/patches/gpsstatus2/ads/DisableAdsPatch.kt
```diff
@@ -0,0 +1,27 @@
...
+import app.revanced.patcher.data.BytecodeContext
+import app.revanced.patcher.extensions.InstructionExtensions.removeInstructions
+import app.revanced.patcher.patch.BytecodePatch
+import app.revanced.patcher.patch.annotation.CompatiblePackage
+import app.revanced.patcher.patch.annotation.Patch
+import app.revanced.patches.gpsstatus2.ads.fingerprints.PlaceAdFingerprint
+import app.revanced.util.exception
+
+@Patch(
+    name = "Disable Ads",
```
oSumAtrIX, line 12:
```suggestion
    name = "Disable ads",
```

src/main/kotlin/app/revanced/patches/gpsstatus2/ads/fingerprints/PlaceAdFingerprint.kt
programminghoch10: Yes this string is present in many methods.

programminghoch10: I know the pattern could be shorter, i just wanted to be extra sure it is the correct method since i didnt find any other unique properties of this method.

programminghoch10: Okay, will look into it.
Is there any way to check if a fingerprint matches only once?

oSumAtrIX: > i just wanted to be extra sure

It could break if the pattern is too big and the app updates, so I recommend to shorten it down, add other attributes like access flags the return type or parameters or fingerprint a better method that references this one. This way you can reach the method this fingerprint matches via a reference in that other method.

programminghoch10:
Thanks for all your feedback,
I think I don't have the time to complete this at the moment.
This was a novelty thing anyways since a normal adblocker already removes the ads, I just wanted the replacement advertisement "Get PRO version" also gone.
I'll reopen when I find the time to do this properly.

PR #3448, opened 2024-07-17
feat: Add `Disable analytics and telemetry` patches

patchink0:
Added generic patches to disable multiple analytics and telemetry SDKs: ComScore, AppsFlyer, MoEngage, Segment, Statsig, Facebook SDK, Google Analytics, Firebase Crashlytics, Firebase Logging

Those are all the SDKs included in the SoundCloud app, and the patches were developed and tested on it but should work for any app.

This greatly increases privacy but also performance, more specifically regarding the startup time. SoundCloud initializes and opens up noticeably faster using those patches.


oSumAtrIX (changes requested):
Looks good on first sight. Once the couple style/refactor issues are addressed it lgtm. I do not know if each patch works here as expected, were they all been tested? The readme of the repository should be changed in this pr as well to include privacy patches as a feature, given this pr introduces pretty much all the common ones.

oSumAtrIX:
That sounds like a good idea indeed.

src/main/kotlin/app/revanced/patches/all/analytics/UniversalPrivacyPatch.kt
```diff
@@ -0,0 +1,49 @@
...
+            description = "Which SDK to target",
+            required = true
+        )
+    }
+
+
+    override fun execute(context: BytecodeContext) {
+        subPatchesOptionsEvaluated.forEach {
+            if (it.value.value == true){
+                Logger.getLogger(this::class.java.name).warning("Applying patch to disable ${it.key.first}")
```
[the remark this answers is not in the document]
oSumAtrIX: Yes, info works here.

src/main/kotlin/app/revanced/patches/all/analytics/appsflyer/fingerprints/AppsFlyerInitFingerprint.kt
```diff
@@ -0,0 +1,14 @@
+package app.revanced.patches.all.analytics.appsflyer.fingerprints
+
+import app.revanced.patcher.extensions.or
+import app.revanced.patcher.fingerprint.MethodFingerprint
+import com.android.tools.smali.dexlib2.AccessFlags
+
+object AppsFlyerInitFingerprint : MethodFingerprint(
```
oSumAtrIX, line 7: Make sure fingerprints are internal

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/all/privacy/UniversalResourcePrivacyPatch.kt
```diff
@@ -0,0 +1,109 @@
...
+import app.revanced.patcher.patch.PatchException
+import app.revanced.patcher.patch.ResourcePatch
+import app.revanced.patcher.patch.annotation.Patch
+import app.revanced.patcher.patch.options.PatchOption.PatchExtensions.booleanPatchOption
+import app.revanced.patches.shared.resource.AndroidManifest
+import java.util.logging.Logger
+
+@Patch(
+    name = "Disable more privacy invasive components",
+    description = "Disables multiple embedded analytics and telemetry SDKs by modifying the app's manifest. Please note this can break some apps.",
```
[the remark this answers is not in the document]
oSumAtrIX: >  can a resource patch be included or excluded at runtime?

ReVanced Patcher checks all added patches and their dependencies before patches are ran so it can decode the resources before any patches are executed. Otherwise while patches run, there would be a delay for the first resource patch running.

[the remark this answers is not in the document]
oSumAtrIX: Options are instance members of a patch instance. When a patch is instantiated its dependencies are set. So there is no way for an option to decide weather a patch depends on another. The situation is the following:

Either we have one patch which also employs a resource patch. One patch that does not employ resource patches at all. One resource and one bytecode patch separately or each privacy patch is its own, so multiple bytecode and resource patches.

The initial idea was that we combine all patches so that the patch handles all kinds of analytics patches itself (or via patch options) but if we want to make it work on apps where resource patches fail, we have to move resources patches to a different patch which breaks our initial idea. So either its this or apps which would otherwise just need bytecode patches could potentially fail due to resource patches.

README.md
```diff
@@ -77,6 +77,7 @@ Some of the features the patches provide are:
 * 🚫 **Block ads**: Say goodbye to ads
 * ⭐ **Customize your app**: Personalize the appearance of apps with various layouts and themes
 * 🪄 **Add new features**: Extend the functionality of apps with lots of new features
+* 📡 **Enhance your privacy**: Disable embedded trackers and privacy invasive components
```
oSumAtrIX, line 80:
```suggestion
* 📡 **Protect your privacy**: Disable embedded trackers and privacy-invasive components
```

oSumAtrIX:
I am locally refactoring this PR

src/main/kotlin/app/revanced/patches/all/privacy/DisableAnalyticsAndTelemetryThroughManifestPatch.kt
```diff
@@ -0,0 +1,88 @@
...
+    private const val META_DATA_TAG = "meta-data"
+    private const val NAME_ATTRIBUTE = "android:name"
+    private const val VALUE_ATTRIBUTE = "android:value"
+
+    private fun Document.addMetadata(vararg metaDataNodes: Pair<String, String>) =
+        metaDataNodes.forEach { (nodeName, nodeValue) ->
+            val applicationNode = getNode("application") as Element
+            applicationNode.getElementsByTagName(META_DATA_TAG)
+                .asSequence()
+                .first { it.attributes.getNamedItem(NAME_ATTRIBUTE).nodeValue == nodeName }
```
oSumAtrIX, line 85: I replaced firstOrNull with first here when refactoring. Correct this if it should be firstOrNull. If it should, then the try captch block above isn't necessary which was present before I refactored.

src/main/kotlin/app/revanced/patches/all/privacy/DisableAnalyticsAndTelemetryThroughManifestPatch.kt
```diff
@@ -0,0 +1,88 @@
...
+@Patch(
+    name = "Disable analytics and telemetry through manifest",
+    description = "Patches the manifest to disable analytics and telemetry services.",
+    use = false,
+)
+@Suppress("unused")
+object DisableAnalyticsAndTelemetryThroughManifestPatch : ResourcePatch() {
+    private val logger = Logger.getLogger(this::class.java.name)
+
+    private val disableFirebaseTelemetry by option("Firebase telemetry")
```
oSumAtrIX, line 22: I merged the two firebase patches to one.

src/main/kotlin/app/revanced/patches/all/privacy/DisableAnalyticsAndTelemetryThroughManifestPatch.kt
```diff
@@ -0,0 +1,88 @@
...
+    private val disableFacebookSdk by option("Facebook SDK")
+    private val disableGoogleAnalytics by option("Google Analytics")
```
oSumAtrIX, line 24: GAnalytics patch also exists in the bytecode patch, why is it present here as well? Also, can't the firebase & facebook patches be converted to bytecode?

oSumAtrIX (changes requested):
Please test the refactoring changes & my review. Otherwise LGTM

cryeprecision:
@patchink0 I'd be willing to fork your fork and try to get this merged if you don't have the time for this at the moment.

PR #3455, opened 2024-07-18
fix(Instagram - Hide ads):  Restore compatibility with latest version by fixing fingerprint

tymmesyde:
Updated the fingerprint of the `Hide ads` patch for the `340.0.0.22.109` version
Should i lock this patch to a specific app version?


oSumAtrIX:
I'd give the fingerprint a second try. Looks like it should survive fine.

PR #3462, opened 2024-07-20
refactor: Bump ReVanced Patcher & merge integrations by using ReVanced Patches Gradle plugin

oSumAtrIX:
# About

- Bumps ReVanced Patcher
- Uses ReVanced Patches Gradle plugin
- Merges integrations

# Todo

- [x] Merge commit history of ReVanced Integrations
- [x] Bump dependencies to stable (Patcher, lib)
- [ ] Sync with https://github.com/ReVanced/revanced-manager/pull/2242
- [ ] Sync with https://github.com/ReVanced/revanced-api (already released but needs to deployed on our server)
- [ ] Test patches
- [x] Fix consistency (top-level `EXTENSION_CLASS_DESCRIPTOR` or local `extensionClassDescriptor`, for example) (Should be converted to extensionClassDescriptor eventually)

# Merge instructions

- This PR should be merged as two commits:
   1. A commit that highlights the ReVanced Patcher bump.
   2. A commit that highlights the merge of integrations

The first commit may have residues from the second commit.


oSumAtrIX:
Please keep in mind I am currently in a long merge conflict that I am resolving since a couple days, so before you push anything wait for that to be done. 

> The patches.json file in the PR is outdated and could be deleted

ReVanced Manager uses this file. The PR to update it to the latest patcher version also migrates it to API v3 so then we can delete the JSON.



> Unrelated to this PR, but if using `--install` without specifying a device serial number it could use the old behavior of picking the first (or only) device available. It makes installing to different devices a lot simpler.

I was under the impression that this is the current behavior. I'll check again.

> Nearly all patches with integrations code have a java file with an identical name:
> 
> `HideAutoplayButtonPatch.kt` `HideAutoplayButtonPatch.java`
> 
> Maybe the integration patch files should drop the `Patch` suffix? Or use some other name so it's unique from the patch files?

Extension projects currently are full blown Android projects. This comes with initial build time issues. Ideally we can find an internal API of the Android Gradle plugin that let's us compile the DEX only instead of the entire APK. This would cut build times by basically 99% since resources don't have to be built for example or no APK has to be packaged. Since the build system is now moved into a plugin this also means we can find such an improvement and then others just have to bump their plugin to get this new build system.

Regarding the file name, you can search in a scope in IntelliJ IDEA, for example in the patches project. This way when you search for the name, the extension class won't be shown. Besides, if you search for a part of the patch name without the "Patch" part, it'll still match both the extension and patch class

patches/build.gradle.kts
```diff
@@ -1,5 +1,8 @@
 group = "app.revanced"
 
+java.sourceCompatibility = JavaVersion.VERSION_11
```
oSumAtrIX, line 3: These should be moved to the patches plugin. I recall setting them there anyway, but I may be wrong

oSumAtrIX: @LisoUseInAIKyrios 

The plugin does indeed set the versions already, so the latest commit is not necessary and can be reverted:

image

oSumAtrIX:
> Unrelated to this PR, but if using `--install` without specifying a device serial number it could use the old behavior of picking the first (or only) device available. It makes installing to different devices a lot simpler.

CLI currently does this:

image

So if you specify `-i` it will use the fallback value and be non null:

image

oSumAtrIX:
> I don't know if the generate patch files gradle task is still relevant but it throws a NPE without any explanation what went wrong:
> 
> ```
>  ./gradlew generatePatchesFiles
> 
> > Task :patches:generatePatchesFiles FAILED
> Exception in thread "main" java.lang.NullPointerException
>         at app.revanced.generator.MainKt.main(Main.kt:7)
>         at app.revanced.generator.MainKt.main(Main.kt)
> ```
> 
> The `patches.json` file in the PR is outdated and could be deleted.

https://github.com/ReVanced/revanced-manager/pull/2242 now makes the Manager use the latest patcher and seems to work as expected. No more integrations as well. It also gets rid of the dependency to `patches.json` therefore this concept has now been removed.

oSumAtrIX:
@LisoUseInAIKyrios This PR is technically ready apart from the TODOs. What I am contemplating is whether we really want to merge the history of integrations into the patches. How it would work is that the integrations commit would have a separate root commit (possible in Git), so when we merge this PR to dev, it would contain all the history of integrations "rewritten" under the extension path. Though this may be a bit weird, and it's rewriting history anyway, it may not be something we really have to do. What do you think?

oSumAtrIX:
Usually, our reports are archived and, after a while, deleted. Then we should merge the history

oSumAtrIX:
Fixed with https://github.com/ReVanced/revanced-cli/commit/5f952f35f5cb388b6509b2b4d905b8143ebc7996

oSumAtrIX:
Shortly, we will make an announcement for devs on our Discord to test these changes using https://github.com/ReVanced/revanced-manager/actions/runs/11282128151/artifacts/2042645943 and https://github.com/ReVanced/revanced-patches-patcher-v20 so that before the release for everyone we can fix some bugs.

oSumAtrIX:
> The patch option delegate is throwing a NPE somewhere.

This is odd. The option is instantiated in a top-level property:

image

It is added to the user-facing patch:

image

The option value is delegated to a property in a dependency patch:

image

The value of the option is initially null, but I am under the impression that delegating null is allowed.

oSumAtrIX:
> The patch option delegate is throwing a NPE somewhere.

Fixed with [`057b749` (#3462)](https://github.com/ReVanced/revanced-patches/pull/3462/commits/057b749d3dd20319e24f76fcbdd53c020d154525):

image

The issue was caused by first initializing the resource patch that delegates the option value but the option itself was not initialized yet, so it was accessing a field from a null reference. Kotlin's null checker analysis falls short here.

oSumAtrIX:
I assume you are referring to the options.json file.

The JSON has been removed. It was not an elegant design for a CLI application. You can save your options into a text file or a script and pipe them into CLI, emulating the same behaviour as the options.json file.

Examples:

- `cat youtube-patch | xargs java -jar revanced-cli.jar`
- `cat dark-theme | xargs java -jar revanced-cli.jar patch -p patches.rvp app.apk`
- `cat my-selection | xargs java -jar revanced-cli.jar patch -p patches.rvp app.apk`
- `cat osumatrix-patches | xargs java -jar revanced-cli.jar patch app.apk`
- `cat multiple-rvps-some-app | xargs java -jar revanced-cli.jar`

If you mean the patches.json file, this concept only existed for a specific purpose, which is no longer. Previously, it was added for ReVanced API, but now ReVanced API uses download, verify the signature and read the patches using ReVanced Patches by itself (https://api.revanced.app/v3/patches/latest/list). This endpoint is specifically useful for our website, so the route may eventually be moved to a website-specific route or API.

oSumAtrIX:
This is ready (Apart from merging integration histories, I'll give that a shot in a few minutes). ReVanced Manager is currently being tested for bugs. Besides that, the new patcher also fixes the known issues with OOM so we can close them.

oSumAtrIX:
The project is going to be converted to KMP anyway since it's targeting multiple platforms

oSumAtrIX:
Sure

oSumAtrIX:
@LisoUseInAIKyrios This PR is now ready for review/mergeing. Before we merge I will try to force push the integrations history into this branch and merge by rebasing. Since there are many objects the repository will grew a bit on size due to stuff like the Gradle wrapper being committed to. I could rewrite history in a way where only commits to the source files are included. This skews true history, but avoids bloating the repo with commits that will be forgotten anyways in a while. What do you say

oSumAtrIX:
Merging this means all other current PRs need to be rebased and resolve merge conflicts. I can take up on some. Since sources are moved its simply a matter of rebasing, and during merge conflict resolution copy over the old source files to the new place and apply the necessary changes.

oSumAtrIX:
> Build failed with out of memory. Locally I had to increase gradle memory argument as recently it had been complaining about running low.

Odd. Maybe environmental specific. Gradle orchestrates a parallelized build for all subprojects such as extensions and the patches project.

oSumAtrIX:
I have merged both histories now:

image

image

The integration commits are rewritten only to include changes to the source files:

image

The rewrite moved all commits to the source files to a directory "intregrations". This allowed for a merge commit without conflicts into the patches repo. The merge was stopped with `--no-commit` and the commit that converted the integrations to an extension was applied to the pending merge. The integrations directory was deleted, the extensions folder created, and the merge finished:

image

oSumAtrIX:
I was thinking of getting rid of the structure entirely and just structure by package name. All patch files would be next to each other. Fingerprint files could be named the same way as the corresponding patch too (HideAdsFingerprints.kt for example). I think this is easier to read.

PR #3468, opened 2024-07-24
fix(YouTube - Client Spoof): Unlock more qualities by spoofing the iOS client user agent

zainarbani:
[Integration changes](https://github.com/ReVanced/revanced-integrations/pull/668)


zainarbani:
> If this fixes higher resolutions for all users/devices, the spoof summary disclaimer can be updated to remove the limitation.

I don't have another devices to test the patch, should i push the limitation update?

oSumAtrIX:
I'd say before we remove the warning and add it back, we can keep it, and remove it once we confirm it works fine

0xrxL:
Background playback for livestream is still bugged, despite this change. What a sin.

oSumAtrIX (approved):
I have pushed some refactoring changes. Please review if the patch still works as expected. Apart from that lgtm.

PR #3480, opened 2024-07-28
fix(YouTube - Spoof client): Fix tracking history on brand accounts

oSumAtrIX:
RE: https://github.com/ReVanced/revanced-integrations/pull/669

### DIFF
--- a/src/main/kotlin/app/revanced/patches/youtube/misc/fix/playback/SpoofClientPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/youtube/misc/fix/playback/SpoofClientPatch.kt
@@ -83,6 +83,9 @@ object SpoofClientPatch : BytecodePatch(
 
         // Player speed menu item.
         CreatePlaybackSpeedMenuItemFingerprint,
+
+        // Watch history.
+        GetTrackingUriFingerprint,
     ),
 ) {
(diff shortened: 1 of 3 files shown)


n1cKz1:
I just patched dev and tested briefly, apparently history still isn't tracking for me

oSumAtrIX:
Use a network debugger with the network debugging patch and check if the patch has successfully changed the request urls authority

n1cKz1:
Will do later once I get home.

I do use adguard but the patched app is whitelisted. History tracks fine when spoofing is off.

n1cKz1:
> check if the patch has successfully changed the request urls authority

I tested with spoof client on and off to check for any differences in the traffic and there wasn't any. Double checked that AdGuard isn't at fault, history also works with VR spoofing as expected. 

Luckily for some reason I'm still unaffected by the playback issue so I'll keep spoofing off for now.

oSumAtrIX:
Which authority was present?

n1cKz1:
These are some requests that appear when opening videos with iOS spoofing:

image
image

What am I supposed to look for?

oSumAtrIX:
You are interested in the watch history requests. So those that have api/stats in their path.

n1cKz1:
Spoofing on:
image

Spoofing off:
image

PR #3504, opened 2024-08-04
fix(YouTube - Spoof client): Restore livestream audio only playback with iOS spoofing

LisoUseInAIKyrios:
Fixes background audio only playback of livestreams when spoofing to iOS

[Integration changes](https://github.com/ReVanced/revanced-integrations/pull/673)


0xrxL:
So in theory, once the history bug will be finally fixed, TestSuite spoof will be removed?

inotia00:
> The fingerprints here need some adjusting, as this doesn't work on all old versions.

See the fingerprint in [this commit](https://github.com/inotia00/revanced-patches/commit/dde5331ba949ed2655ae168a6bc2485ebec197e9#diff-0dab1971be6abd6e3e4fc507f8756151e726791ca3bb885c26bfbc98b142a145)
(it's the same as 'PlayerResponseModelAudioOnlyPlaybackFingerprint' in this PR)

The fix also requires a 'BackgroundPlaybackPatch'
(See [this comments](https://github.com/inotia00/revanced-patches/blob/dde5331ba949ed2655ae168a6bc2485ebec197e9/src/main/kotlin/app/revanced/patches/youtube/utils/fix/client/SpoofClientPatch.kt#L399))

0xrxL:
> The fingerprints here need some adjusting, as this doesn't work on all old versions.

You can resolve it by find these two strings:

- ```Failed to start foreground priority player Service due to Android S+ restrictions.```
- ```About to stop background service while in a pending state.```

after these you'll found a synchronized method, that contain a reference to the targeted boolean method to patch.

inotia00:
> > The fix also requires a 'BackgroundPlaybackPatch'
> 
> Is background playback patch really needed? With a premium account and excluding background playback patch, livestream audio only background play is working for me. And if a user without premium excludeds background play then playback won't work outside the app so it's not a concern there.

The current PR always enables background playback if the Spoof client patch is turned on, even if the `Remove background playback restrictions` patch is excluded (even for non-Premium accounts)

The unintended behavior I'm talking about is:
1. Exclude `Remove background playback restrictions` patch, and include `GmsCore support` patch
3. `Spoof client` patch always enables background playback
4. In non-Premium accounts, if the `Remove background playback restrictions` patch is not included, background playback settings are not added, so there is no way to turn off background playback

So I thought it would be appropriate to add the `BackgroundPlaybackPatch` patch as a dependency

src/main/kotlin/app/revanced/patches/youtube/misc/fix/playback/fingerprints/PlayerResponseModelBackgroundAudioPlaybackFingerprint.kt
```diff
@@ -0,0 +1,37 @@
...
+import app.revanced.patcher.extensions.or
+import app.revanced.patcher.fingerprint.MethodFingerprint
+import com.android.tools.smali.dexlib2.AccessFlags
+import com.android.tools.smali.dexlib2.Opcode
+
+internal object PlayerResponseModelBackgroundAudioPlaybackFingerprint : MethodFingerprint(
+    returnType = "Z",
+    accessFlags = AccessFlags.PUBLIC or AccessFlags.STATIC,
+    parameters = listOf("Lcom/google/android/libraries/youtube/innertube/model/player/PlayerResponseModel;"),
+    opcodes = listOf(
```
oSumAtrIX, line 12: Is there any reason this fingerprint is so big?

inotia00: from YouTube 17.34.36 to YouTube 19.31.xx, the target method has the same structure
(accessflags, parameters, return type, opcode pattern are always the same)

I don't think the fingerprint will break easily

oSumAtrIX: The opcode pattern should be reduced if possible.

oSumAtrIX (approved):

ProTechnomancy:
Is playback history broken again? Or is it just me? I'm on android.

I've updated and reinstalled several times and no good

PR #3518, opened 2024-08-08
feat(SCB Easy): Remove broken `Remove debugging detection` patch

### LINKED ISSUE #3517: bug: Removal of SCB Easy Patch
### Bug description

App: `com.scb.phone`

Usually, broken patches will be reported and the developers could fix the patch so that it could work again or constrain it to the last known working version.

But from my research, I find it's too complicated to create a patch because the application is heavily guarded with lots of anti-tampering tools. The application isn't available on APKMirror, you will need to pull it from [Google Play Store](https://play.google.com/store/apps/details?id=com.scb.phone) or [Huawei AppGallary](https://appgallery.huawei.com/#/app/C100999491).

From what I could find, the application is protected by [Guardsquare](https://www.guardsquare.com/) and is dominantly an Android (likely Java) app with Flutter for some of its in-app features. 

The application employed multiple obfuscation and detection tactics to prevent tampering or suspicious activity:
* Root detection
* Android Device Bridge detection
* Accessibility-enabled list check for suspicious application
* VPN service detection
* SafetyNet Attestation
* Threat monitoring service (Provided Guardsquare [ThreatCast](https://www.guardsquare.com/threatcast-mobile-threat-defense))
* Anti-hooking/modification (Likely provided by Guardsquare - mostly will crash with "divide by zero" or "negative one" error)
* Approved app install vendor detection (Known Trusted Installers are: Google Play Store and Huawei AppGallary)
* Bootloader status check[^1]

[^1]: (This is speculation as I haven't tested this, it might have been a PlayIntegrity check but I couldn't get this app to work without a Strong PlayIntegrity check - It also might have been Huawei [SysIntegrity](https://developer.huawei.com/consumer/en/doc/Security-Guides/dysintegritydevelopment-0000001050156331) from their [Safety Detect](https://developer.huawei.com/consumer/en/hms/huawei-safetydetectkit/) suite which checked the TEE) - (Environment: Samsung M23 *SM-M236B* - Android 14 - One UI 6)

So I would suggest removing the SCB Easy patch entirely as it's not worth it to bypass all of the detection features with a codebase that's likely going to be changing with every release.

### Error logs

_No response_

### Solution

_No response_

### Additional context

_No response_

### Acknowledgements

- [X] I have checked all open and closed bug reports and this is not a duplicate.
- [X] I have chosen an appropriate title.
- [X] All requested information has been provided properly.

validcube:
Close #3517


oSumAtrIX:
Please commit the api changes by running the apiDump task. Additionally deprecate the class and remove the name instead of removing the classes.

validcube:
> Please commit the api changes by running the apiDump task. Additionally deprecate the class and remove the name instead of removing the classes.

Done:tm:

PR #3543, opened 2024-08-15
fix(YouTube - GmsCore Support): Use correct patch offset

LisoUseInAIKyrios:
Fixes the fix used for Google Photos integration


src/main/kotlin/app/revanced/patches/shared/misc/gms/BaseGmsCoreSupportPatch.kt
```diff
@@ -115,15 +115,16 @@ abstract class BaseGmsCoreSupportPatch(
...
+            var setContextIndex = indexOfFirstInstruction {
+                val reference = getReference<MethodReference>() ?: return@indexOfFirstInstruction false
 
                 reference.toString() == "Lapp/revanced/integrations/shared/Utils;->setContext(Landroid/content/Context;)V"
             }
 
             // Add after setContext call, because this patch needs the context.
             addInstructions(
-                setContextIndex + 1,
+                if (setContextIndex < 0) 0 else setContextIndex + 1,
```
oSumAtrIX, line 127: I thought of this too, but why is there no setContext for YouTube in the main activity?

Issue #3208, opened 2024-05-21
bug(YouTube): `Spoof client` patch - various side effects


kazimmt:
### Bug description

- [ ] The playback speed option is missing from the flyout menu. #3203
- [ ] Watch history does not work with a brand account. #3200
- [ ] Live videos are only available in 480p resolution.
- [ ] Subtitle do not work by default.
- [ ] Player gestures do not work by default.

### Additional context

Track all side effects in one place.

### Acknowledgements

- [X] This issue is not a duplicate of an existing bug report.
- [X] I have chosen an appropriate title.
- [X] All requested information has been provided properly.

0xrxL:
I think that iOS spoof should be the first thing to enable, because TestSuite it's too limited.

oSumAtrIX:
I don't know if iOS supports HDR playback at all, but if it does, one of the request parameters may be responsible. For example currently the device type is still Android, things like that could be cause of the problem.

0xrxL:
> I don't know if iOS supports HDR playback at all, but if it does, one of the request parameters may be responsible. For example currently the device type is still Android, things like that could be cause of the problem.

It seems: https://www.iphoneincanada.ca/2018/05/01/youtube-ios-hdr-iphone-x/

...along with 60fps. Therefore the spoof should be improved, like you said.

eldepor:
Hi. I don't know if this is spoof client patch side effect, but when I open a clip, YT plays the original video form the beginning, not the clip fragment.

inotia00:
> It seems: https://www.iphoneincanada.ca/2018/05/01/youtube-ios-hdr-iphone-x/
> 
> ...along with 60fps. Therefore the spoof should be improved, like you said.

I was looking through the NewPipeExtractor repository and found a hint about this:
https://github.com/TeamNewPipe/NewPipeExtractor/issues/680#issuecomment-1002724558

You can play 60fps video by spoofing the model name to iOS device with the `modelName` field of the `InnertubeContext$ClientInfo` class.

In YouTube v19.11.43, the `modelName` field name is `w`:
```
const-string v1, "iPhone15,4"
iput-object v1, v0, Lcom/google/protos/youtube/api/innertube/InnertubeContext$ClientInfo;->w:Ljava/lang/String;
```


Also, checking the [response of clientName](https://github.com/zerodytrash/YouTube-Internal-Clients?tab=readme-ov-file#clients) can be helpful in finding side effects that have not yet been discovered.

[...]

PR #3582, opened 2024-08-24
fix(YouTube): Fix issues related to playback by replace streaming data

zainarbani:
[Integration changes](https://github.com/ReVanced/revanced-integrations/pull/680)

Todo:
- [x] Refactor
- [x] Test if the URLs really work as is (because usually they need post processing like in NewPipe)
- [x] Check if properly optimized (make request as early as possible in parallel)
- [x] Test what can be ticked in #3208  (edit: it appears that all issues are fixed)
- [x] Using iOS client as streaming data sources (more playable videos than VR and other).


YT-Advanced:
Uhm so what is the different

oSumAtrIX:
Patches the response

zainarbani:
Another backup plan maybe 🤷‍♂️

0xrxL:
> Another backup plan maybe 🤷‍♂️

This approach can fix at least one of still existing issues with client spoofing? For example brand account history or HDR with vp9 codec?

oSumAtrIX:
The VR client supports hdr streams. The more important question is, if the streams work because usually they need additional post processing as seen by NewPipeExtractor for example. YouTube may apply the post processing from the original request though which could work without any further changes

zainarbani:
@0xrxL HDR vp9.2 works, though i don't remember if vr client has av01 format

zainarbani:
HDR on VR client could be fixed by replacing player config with android one, but its not the only things that missing from vr player response fields. Also VR client lack of kids videos (maybe other as well).

As for iOS client, i've tried to replace the player config but video playback doesn't work which im not sure what happened.

Pamilg9:
Currently, NewPipe cannot play 18+ content. This PR can cause problems

zainarbani:
@Pamilg9 Its fixed by adding Auth headers.

zainarbani:
> If there are any issues with replacing streams for any specific video or situation, then it should be possible to fall back to using the existing client type spoofing approach.

Agreed, i think the worst scenario is we got zero streaming data from client list, or streaming data simply unplayable at all.

oSumAtrIX:
The stream spoofing should ideally replace the client spoof patch entirely. Patching the request comes with issues fixed by patching the response. The only concern is that the streams returned by our custom request may not work because the custom request is not done properly.

zainarbani:
Implementing pre-fetch streaming data maybe a good idea, faster load while scrolling shorts/general videos. but i think VideoIdHook patch no longer detect the next videoId with /initplayback being blocked. There's a way to invalidate /initplayback reponse data instead blocking it, but i don't quite remember if this methods allow videoId hook patch work normally again.

zainarbani:
@oSumAtrIX Do we neeed to follow [this way](https://github.com/LuanRT/YouTube.js/pull/708)? adding poToken on innertube body and uri params

YT-Advanced:
> @oSumAtrIX Do we neeed to follow [this way](https://github.com/LuanRT/YouTube.js/pull/708)? adding poToken on innertube body and uri params

If u want to add another client, like `ANDROID` or `WEB`, u have to add `potoken`. Otherwise, it is not necessary

Pamilg9:
NewPipe has playback issues. Still not fixed

zainarbani:
> NewPipe has playback issues. Still not fixed

We're not using NewPipe atm, and what kind of issue?

oSumAtrIX:
What if you go incognito?

oSumAtrIX:
How about regular videos?

zainarbani:
incognito means no authorization?

oSumAtrIX:
I was asking because of:

> But both of these issues are because NewPipe does not use any YT user authentication. Those issues do not apply here.

This applies to incognito so I was suspecting issues here as well.

oSumAtrIX:
Is the request itself failing or YouTube using the URLs? If it's the former, the request needs to be adjusted. I know that the user agent plays a role, so maybe that could be it, or other body parameters

zainarbani:
https://github.com/user-attachments/assets/a10c990a-d79c-4ea2-a240-c5db55ce0ad8

Strange, its working on my side

zainarbani:
Yeah its VR

zainarbani:
Can you take a look at /player request ?

zainarbani:
I don't  have another devices nor emulator atm.
@0xrxL can you test this? dev build available on my fork

oSumAtrIX:
The reasons the streams are fetched before the response is for speed. While the original request is made, in parallel the custom request can be made. This shaves off some time which otherwise would be impossible if the custom request is made in sequence after the response

oSumAtrIX:
I am not speaking for the client spoof patch here. Like I suggested before, ideally it can be removed. The smaller a patch the better for future maintenance so i am speaking in terms of just this patch for the future to fix playback. So while YouTube does it's real /get_watch request which returns broken streams in the streaming data array, a parallel request is done to replace the streams once the response of the original request is available. Im theory, if both requests take about the same amount of time, there should be no visible playback delay, except if bandwidth bottlenecks (which is unlikely for requests of such small size)

zainarbani:
Current implementation is pretty fast for me, the custom request also done before original /player request which is enough.

zainarbani:
We could also disable /initplayback blocking to avoid multiple switching between /initplayback > get_watch > player.
/initplayback response also contain media data (those ~10 sec working videos before buffering),  but this media data also the one that cause stuttering/jitter on initial shorts playback.

oSumAtrIX:
Blocking is done by the client spoof patch which should be removed in favour of this patch if it works reliable

zainarbani:
Well, im not sure about remove or replace the current client spoof method with this patch.
Speaking about reliability, the only drawback for me is the need to removing ios client /videoplayback post body which is kinda odd. Vr should be fine though, also we can add more fallback if needed.

oSumAtrIX:
Why would the current client spoof be necessary with this PR?

oSumAtrIX:
The patch should ideally work, and a backup solution is not needed. If YouTube changes something that needs the current patch, we can revert this PR and bring it back, but unless strictly necessary, I would prefer a clean and minimal patch instead of one that might require backups or fallbacks and similar. This approach does not seem like it would break before the current approach either, so using the current approach as a backup might also likely not be necessary in the first place.

zainarbani:
@MarcaDian Have you sync the integrations? Also clear rv manager data sometimes help

zainarbani:
> > @MarcaDian Have you sync the integrations? Also clear rv manager data sometimes help
> 
> I compiled patches and integrations from this PR on 2 different PCs, the result is the same. the first time I patched through cli, the second time through manager compose. Is everything working for you?

Yep, your error indicates that the integrations wasn't synced.
You can try my fork build for testing

0xrxL:
> > > @MarcaDian Have you sync the integrations? Also clear rv manager data sometimes help
> > 
> > 
> > I compiled patches and integrations from this PR on 2 different PCs, the result is the same. the first time I patched through cli, the second time through manager compose. Is everything working for you?
> 
> Yep, your error indicates that the integrations wasn't synced. You can try my fork build for testing

Tried it on my phone too, and everything is fixed (along with VP9-HDR). Good thing.

YT-Advanced:
Since this PRs only make changes to streamingData, I think we can use `WEB`, `MWEB` or `TVHTML5` as Client.

oSumAtrIX:
Web requires a BotGuard token

0xrxL:
> It seems to be working fine, what is the delay for the dev release?

A better set of clients to fetch stream data, probably. @zainarbani suggested it on tha last comment.

Anyway ```ExperimentalUrlRequest``` is deprecated on AS, should be fixed I believe.

0xrxL:
> > > It seems to be working fine, what is the delay for the dev release?
> > 
> > 
> > A better set of clients to fetch stream data, probably. @zainarbani suggested it on his last comment.
> > Anyway `ExperimentalUrlRequest` is deprecated on AS, should be fixed I believe.
> 
> Android VR and iOS both have no apparent limitations, and 2 clients is the same number as the current client type spoofing has.
> 
> It doesn't matter if `ExperimentalUrlRequest` is deprecated or not. It's been deprecated since 2022 yet YT keeps using it, and even if that changes the patch will be adjusted to whatever the replacement is. Edit: No experimental features are used by Integrations, and the parent `UrlRequest` class type can be used.

Thanks for explanation, anyway I have another question:

Compared to the numerous attempts made by Google to block video playback, I noticed that this patch has the particularity of changing the data stream source (and the relative query parameters) quite quickly. How useful do you think this patch will be in the future? Because at the moment I don't see any way for Google to block it again...I believe.

It's worth keeping it as permanent patch (without toggling it)?

oSumAtrIX:
The moment the request asks for a PoToken, the patch will break.

0xrxL:
> The moment the request asks for a PoToken, the patch will break.

This practically means implement PoToken in every client type. It's weird that Google hasn't already do this.

oSumAtrIX:
Is there any reason it is currently falling back? The replacement of the response should work for /get_watch

0xrxL:
> Is there any reason it is currently falling back? The replacement of the response should work for /get_watch

Give a look to this. Seems that someone successfully managed to generate the PoToken key:

https://github.com/iv-org/invidious/issues/4734#issuecomment-2290075097

YT-Advanced:
> Give a look to this. Seems that someone successfully managed to generate the PoToken key:
> 
> https://github.com/iv-org/invidious/issues/4734#issuecomment-2290075097

PoToken which generated by BotGuard can't be used in Android

0xrxL:
> > Give a look to this. Seems that someone successfully managed to generate the PoToken key:
> > [iv-org/invidious#4734 (comment)](https://github.com/iv-org/invidious/issues/4734#issuecomment-2290075097)
> 
> PoToken which generated by BotGuard can't be used in Android

They talking about BotGuard/DroidGuard (not specifically one of them). 🤷‍♂️

oSumAtrIX:
The webguard token can be used for the token for Android clients (at least the other way round). Currently webguard is being emulated so we'll see how it holds against time.

oSumAtrIX:
I think there's still an issue with Shorts playback. I'll review again shortly

zainarbani:
> I think there's still an issue with Shorts playback. I'll review again shortly

Those glitch issue shouldn't be happen since /initplayback still blocked, or another issue?

0xrxL:
> > I think there's still an issue with Shorts playback. I'll review again shortly
> 
> Those glitch issue shouldn't be happen since /initplayback still blocked, or another issue?

I don't know if it's the same issue, but two (rarely) times I had to restart the app due to infinite buffering loading (on both video and shorts).

This could have something to do with the removal of blockGetWatch method?

oSumAtrIX:
The issue i was referring to was about the initial frame being distorted. It's fixed by blocking initplayback but ideally that shouldn't be necessary

zainarbani:
@0xrxL yeah i noticed that as well, also fetch streaming data sometimes failed with resp code 401.

0xrxL:
> @0xrxL yeah i noticed that as well, also fetch streaming data sometimes failed with resp code 401.

Ok, I left youtube in background for a while and right now i got the same problem. Maybe the cause is the offMainThread function not handled properly?

zainarbani:
Its either /get_watch streaming data processed on somewhere else too or something else caused it.
@LisoUseInAIKyrios Does streaming data cache got expired/deleted wether its used(called) or no?

zainarbani:
Alright, i'll check the logs if this happens again.

zainarbani:
Due to iOS HLS implementation?

0xrxL:
What's still missing on this PR to be merged to dev? The main thread check?

0xrxL:
So it's all technically ready?

src/main/kotlin/app/revanced/patches/youtube/misc/fix/playback/SpoofClientPatch.kt
```diff
@@ -172,228 +148,103 @@ object SpoofClientPatch : BytecodePatch(
...
+                replaceInstruction(
+                    buildRequestIndex,
+                    "invoke-static { v$requestBuilderRegister, v$urlRegister, v${urlRegister + 1} }, " +
+                            "$INTEGRATIONS_CLASS_DESCRIPTOR->" +
+                            "buildRequest(${REQUEST_BUILDER_CLASS_DESCRIPTOR}Ljava/lang/String;Ljava/util/Map;)" +
+                            REQUEST_CLASS_DESCRIPTOR
+                )
+                
+                // Copy request headers for streaming data fetch.
+                addInstruction(newRequestBuilderIndex + 2, "move-object v${urlRegister + 1}, p1")
```
oSumAtrIX, line 171: What's urlRegister + 1?

src/main/kotlin/app/revanced/patches/youtube/misc/fix/playback/SpoofClientPatch.kt
```diff
@@ -172,228 +148,103 @@ object SpoofClientPatch : BytecodePatch(
...
+                    videoDetailsIndex + 1,
+                    """
+                        # Registers is free at this index.
+
+                        invoke-static { }, $INTEGRATIONS_CLASS_DESCRIPTOR->isSpoofingEnabled()Z
+                        move-result v1
+                        if-eqz v1, :disabled
+
+                        # Get video id.
+                        iget-object v1, v0, $videoDetailsClass->c:Ljava/lang/String;
```
oSumAtrIX, line 198: Hardcoded variable name is fine here since these will unlikely change.

src/main/kotlin/app/revanced/patches/youtube/misc/fix/playback/SpoofClientPatch.kt
```diff
@@ -172,228 +148,103 @@ object SpoofClientPatch : BytecodePatch(
...
-                        move-result v$shouldCreateMenuRegister
+                        # Field a: Stream uri.
+                        # Field c: Http method.
+                        # Field d: Post data.
+                        iget-object v1, v0, $definingClass->a:Landroid/net/Uri;
+                        iget v2, v0, $definingClass->c:I
+                        iget-object v3, v0, $definingClass->d:[B
+                        invoke-static { v1, v2, v3 }, $INTEGRATIONS_CLASS_DESCRIPTOR->removeVideoPlaybackPostBody(Landroid/net/Uri;I[B)[B
+                        move-result-object v1
+                        iput-object v1, v0, $definingClass->d:[B
```
oSumAtrIX, line 242: Hardcoded registers could be problematic in case of bumping versions. Are you sure these won't change?

src/main/resources/addresources/values/strings.xml
```diff
@@ -1148,38 +1148,19 @@ This is because Crowdin requires temporarily flattening this file and removing t
...
-            <string name="revanced_spoof_storyboard_summary_on">Storyboard spoofed</string>
-            <string name="revanced_spoof_storyboard_summary_off">Storyboard not spoofed\n\nSide effects include:\n• No ambient mode\n• Seekbar thumbnails are hidden</string>
-            <string name="revanced_spoof_storyboard_timeout">Spoof storyboard temporarily not available (API timed out)</string>
-            <string name="revanced_spoof_storyboard_io_exception">Spoof storyboard temporarily not available: %s</string>
+            <string name="revanced_spoof_streaming_data_title">Spoof client</string>
+            <string name="revanced_spoof_streaming_data_summary_on">Client is spoofed</string>
+            <string name="revanced_spoof_streaming_data_summary_off">Client is not spoofed\n\nVideo playback may not work</string>
+            <string name="revanced_spoof_streaming_data_user_dialog_message">Turning off this setting may cause video playback issues.</string>
+            <string name="revanced_spoof_streaming_data_type_title">Default client</string>
+            <string name="revanced_spoof_streaming_data_ios_force_avc_title">Force AVC (H.264)</string>
```
oSumAtrIX, line 1156: What's the reason for this switch again? VP9 or AVC should be detected automatically (as implemented in integrations). Is this just a toggle for those who experience issues even though the automatic check concluded that they should use VP9? Would the unpatched app not use VP9 too?

src/main/kotlin/app/revanced/patches/youtube/misc/fix/playback/SpoofClientPatch.kt
0xrxL: It still works on 19.36.35 (despite it's a beta).

Check it on my repo if you need of a comparison.

oSumAtrIX: Link the repository

oSumAtrIX: I don't have access to such repository, but i don't want any access to it either. Consider open sourcing it

0xrxL: > > uDrop (the private one where you are already in).
> 
> I don't have access to such repository, but i don't want any access to it either. Consider open sourcing it

I already evaluated this option some days ago, and I think I'll do it soon. I just need to add a way to automatically generate a sign key.

src/main/kotlin/app/revanced/patches/youtube/misc/fix/playback/SpoofClientPatch.kt
zainarbani: This is before old spoof client completely removed (refactored).

src/main/kotlin/app/revanced/patches/youtube/misc/fix/playback/SpoofClientPatch.kt
zainarbani: This is init methods, so basically just field set.

kitadai31:
> The force AVC setting now behaves differently if the device does not have VP9 or AV1 hardware decoding, to better inform the user to use the iOS client type.

Have you confirmed that Android VR uses VP9 regardless of device codecs in Spoof streaming data?

In Spoof **client**, Android VR respects the device codec and uses AVC for devices that don't support VP9.
So Android VR had no problems on such devices in Spoof client.

src/main/resources/addresources/values/strings.xml
oSumAtrIX: So this is not a toggle that is added due to the spoofing patch, but rather a separate feature brought by the spoofing patch as a nice to have

kitadai31:
> The only 'old' device I can try is the Android simulator

Okey. Then I think Android VR is likely to respect the hardware codecs in Spoof streaming data.

src/main/kotlin/app/revanced/patches/youtube/misc/fix/playback/SpoofClientPatch.kt
0xrxL: > > uDrop (the private one where you are already in).
> 
> I don't have access to such repository, but i don't want any access to it either. Consider open sourcing it

uDrop is now open source 👍

src/main/kotlin/app/revanced/patches/youtube/misc/fix/playback/fingerprints/StatsQueryParameterFingerprint.kt
```diff
@@ -1,8 +0,0 @@
-package app.revanced.patches.youtube.misc.fix.playback.fingerprints
-
-import app.revanced.patcher.fingerprint.MethodFingerprint
-
-@Deprecated("Fingerprint is obsolete and will be deleted soon")
-internal object StatsQueryParameterFingerprint : MethodFingerprint(
```
oSumAtrIX, line 6: These are internal and dont need to be deprecated

src/main/kotlin/app/revanced/patches/youtube/misc/fix/playback/SpoofClientPatch.kt
```diff
@@ -81,35 +79,15 @@ import com.android.tools.smali.dexlib2.immutable.ImmutableMethodParameter
 )
 object SpoofClientPatch : BytecodePatch(
```
oSumAtrIX, line 80: The patch should be renamed to what it actually does which is spoofing the playback response. I am not sure what name would fit but something like "ReplaceVideoStreamsPatch" or "FixPlaybackPatch" or "SpoofVideoStreamsPatch". The description can then say something like "Replaces the video streams with working ones to fix video playback otherwise prevented by patching the app."

src/main/kotlin/app/revanced/patches/youtube/misc/fix/playback/SpoofClientPatch.kt
```diff
@@ -172,228 +153,128 @@ object SpoofClientPatch : BytecodePatch(
...
+                val protobufClass = ProtobufClassParseByteBufferFingerprint.resultOrThrow().mutableMethod.definingClass
+
+                val setStreamingDataIndex = result.scanResult.patternScanResult!!.startIndex
+                val playerProtoClass = getInstruction(setStreamingDataIndex + 1)
+                    .getReference<FieldReference>()!!.definingClass
+                val setStreamingDataField = getInstruction(setStreamingDataIndex).getReference<FieldReference>()
+                val getStreamingDataField = getInstructions().find { instruction ->
+                    instruction.opcode == Opcode.IGET_OBJECT &&
+                            instruction.getReference<FieldReference>()?.definingClass == playerProtoClass
+                }?.getReference<FieldReference>() ?: throw PatchException("Could not find getStreamingDataField")
```
oSumAtrIX, line 201: It would be good to prevent such big codeblocks by creating fingerprints with opcode patterns and then using the indices from the patterns.

src/main/kotlin/app/revanced/patches/youtube/misc/fix/playback/SpoofClientPatch.kt
oSumAtrIX: > The advantage of still using Spoof client is everyone already knows what the patch is and why it's needed. Online guides still refer to Spoof client as well. End users don't care what the patch is called or the technical details of how it works, all that matters to them is it fixes playback.

I believe renaming the patch is still the right approach. Using a misleading name isn't ideal, as it doesn't accurately describe what the patch does. ReVanced Manager already has the option to use the recommended patches, which these guides should mention instead of the patch for people who do not know what patches to keep. Patches such as this that are highly dependant on what YouTube does can be renamed, modified, or removed, so any guide mentioning a specific patch name must account for this behavior.

> Technically, the client is still spoofed, but only the client types for the streams are spoofed (not the entire client request). So Spoof client is still valid. The summary could mention it spoofs the client streams.

The patch replaces the response streams by spoofing a different client to **get** those streams, but it doesn’t spoof the actual YouTube request itself. So while the client is spoofed to an extent, the focus is really on the stream data. I agree that `Spoof streaming data` is closer to what the patch does, but the term `streaming data` might be too technical. Something like `Spoof video streams` is more intuitive, or even clearer: `Replace video streams`. Let me know what you think.

oSumAtrIX (approved):

oSumAtrIX:
Thanks :)

PR #3610, opened 2024-09-01
feat: Add `Check environment` patch

oSumAtrIX:
# About

This PR adds a patch that checks and warns the user in case they have downloaded the app as a pre-patched APK or if the APK was not patched and installed using the expected means, such as ReVanced Manager or ReVanced CLI.

# How it works

Several checks are run in sequence. Some checks may exempt others. One check is comparing the hardware of the device used to patch vs what device the app is then launched on. Another checks if the patch & install time is about the same.  And another check verifies the installer source is either CLI or ReVanced Manager.

If the environment check is deemed to have failed, then a warning will be issued at the app launch. The warning will mention the current situation and explain why the warning is being shown by mentioning what kind of checks have failed. The warning has a button to open the source to patch and install an app as intended and a button to dismiss the warning. If the warning is ignored twice it will never be shown again. If the environment check is deemed successful, then the check will be disabled forever.

# Quirks

- Of course, while unlikely, false positives could still occur.  A false positive can occur when you want to share your patched app with friends or family. The warning mentions what kind of checks lead to the warning, giving the user the required context to determine whether they should take the suggested action or ignore it.

[Integration changes.](https://github.com/ReVanced/revanced-integrations/pull/683)

# Todo

- [x] Check if the startup time is unaffected (Maybe run in a background thread?)
- [x] Resolve the remaining todo
- [x] Check that false positives unlikely occur
- [x] Verify the device build info check when installed from ReVanced Manager

Warning


src/main/resources/addresources/values/strings.xml
```diff
@@ -31,6 +31,18 @@ This is because Crowdin requires temporarily flattening this file and removing t
...
...&lt;p&gt;It is strongly recommended to &lt;b&gt;uninstall this version of the app and patch it yourself&lt;/b&gt; to ensure, all patches are intended and secure.&lt;p&gt;The following checks have failed and concluded this app to be pre-patched, obtained from someone else other than you, not correctly functional or &lt;b&gt;harmful and even dangerous&lt;/b&gt; to use:&lt;/p&gt;&lt;p&gt;This warning will only be shown once more, if dismissed, and never again.&lt;p&gt;Uninstall this version of the app and patch it yourself, by opening the offical link below.&lt;/p&gt;&lt;br&gt;&lt;h2&gt;Official links&lt;/h2&gt;&lt;br&gt;&lt;p&gt;ReVanced is always available at &lt;a href=https://revanced.app&gt;revanced.app&lt;/a&gt;.&lt;/p&gt;&lt;br&gt;&lt;ul&gt;&lt;li&gt;&lt;a href=https://revanced.app&gt;Website&lt;/a&gt;&lt;li&gt;&lt;a href=https://revanced.app/discord&gt;Discord&lt;/a&gt;&lt;li&gt;&lt;a href=https://www.reddit.com/r/revancedapp&gt;Reddit&lt;/a&gt;&lt;li&gt;&lt;a href=https://twitter.com/revancedapp&gt;Twitter&lt;/a&gt;&lt;li&gt;&lt;a href=https://t.me/app_revanced&gt;Telegram&lt;/a&gt;&lt;li&gt;&lt;a href=https://www.youtube.com/@ReVanced&gt;YouTube&lt;/a&gt;&lt;/ul&gt;</string>
```
oSumAtrIX, line 39: @LisoUseInAIKyrios This may have some errors (the HTML itself is minified). Using HTML for formatting. Also, the text may be adjusted to be more precise or more correct if necessary. It should not make baseless assumptions (e.g. telling the user that this app is a virus, but instead suggest that it could be)

src/main/kotlin/app/revanced/patches/shared/misc/checks/BaseCheckEnvironmentPatch.kt
```diff
@@ -0,0 +1,161 @@
...
+
+    private fun setPatchInfo() {
+        PatchInfoFingerprint.setClassFields(
+            "PUBLIC_IP_DURING_PATCH" to (publicIp ?: "").encodedAndHashed,
+            "PATCH_TIME" to System.currentTimeMillis().encoded,
+        )
+
+        fun setBuildInfo() {
+            try {
+                Class.forName("android.os.Build")
```
oSumAtrIX, line 62: This check is for patching on platforms other than Android.

src/main/kotlin/app/revanced/patches/shared/misc/checks/BaseCheckEnvironmentPatch.kt
```diff
@@ -0,0 +1,161 @@
...
+import java.io.InputStreamReader
+import java.net.HttpURLConnection
+import java.net.URL
+import java.security.MessageDigest
+import java.util.logging.Logger
+import kotlin.io.encoding.Base64
+import kotlin.io.encoding.ExperimentalEncodingApi
+import kotlin.random.Random
+
+abstract class BaseCheckEnvironmentPatch(
```
oSumAtrIX, line 29: This patch could be universal but:

1. it would need to involve a resource patch or
2. a raw resource patch to decode read the manifest axml

to read the mainactivity from it. But if it were universal, it could be automatically applied to every app.

YT-Advanced:
I don't think IP checking is needed.
And how can we detect if user patched the app via CLI ._.

oSumAtrIX:
Both, the time and ip check will succeed the check. If Android allows, an installation of ReVanced Manager will also let it pass

oSumAtrIX:
@LisoUseInAIKyrios There are a couple of todos left in the code and the PR body. One important test is if everything works on Android. The patch code, for example, uses the URL connection, but it is unclear if it also needs to run in a background thread.

src/main/kotlin/app/revanced/patches/shared/misc/checks/BaseCheckEnvironmentPatch.kt
```diff
@@ -0,0 +1,162 @@
...
+    private val mainActivityOnCreateFingerprint: MethodFingerprint,
+    compatiblePackages: Set<CompatiblePackage>,
+    integrationsPatch: BaseIntegrationsPatch,
+) : BytecodePatch(
+    name = "Check environment",
```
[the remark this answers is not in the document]
oSumAtrIX: Some apps do not have many patches a settings patch. This patch also relies on the integrations patch which means it would create a cyclic dependency if we make integrations depend on it, leaving us with the only option of either making individual patches depend on it, or the base settings patch

src/main/kotlin/app/revanced/patches/shared/misc/checks/BaseCheckEnvironmentPatch.kt
oSumAtrIX: It could be that Android Studio gets confused with the stub module having the same package as Android. I exclusively use IntelliJ IDEA for ReVanced as it has an Android plugin in case I need it for ReVanced Manager or other Android projects. It's also more feature rich and suitable for patches so maybe give it a shot instead of Android Studio.

src/main/resources/addresources/values/strings.xml
```diff
@@ -31,6 +31,15 @@ This is because Crowdin requires temporarily flattening this file and removing t
 -->
 <resources>
     <app id="shared">
+        <patch id="misc.checks.BaseCheckEnvironmentPatch">
+            <string name="revanced_check_environment_failed_title">Checks failed</string>
+            <string name="revanced_check_environment_dialog_open_official_source_button">Open official website</string>
+            <string name="revanced_check_environment_dialog_ignore_button">Ignore</string>
+            <string name="revanced_check_environment_failed_message">&lt;h5>This app does not appear to be patched by you.&lt;/h5>&lt;br>It\'s possible this install will not function correctly or &lt;b>could be harmful or dangerous to use&lt;/b>.&lt;br>&lt;br>The following checks have failed and imply this app is pre-patched or obtained from someone else:&lt;br>&lt;br>&lt;small>%1$s&lt;/small>&lt;br>It is strongly recommended to &lt;b>uninstall this app and patch it yourself&lt;/b> to ensure you are using a validated and secure app.&lt;p>&lt;br>&lt;p>If this warning is ignored it will not be shown again.</string>
```
KobeW50, line 38:
```suggestion
            <string name="revanced_check_environment_failed_message">&lt;h5>This app does not appear to be patched by you.&lt;/h5>&lt;br>This app may not function correctly or &lt;b>may be harmful or dangerous to use&lt;/b>.&lt;br>&lt;br>The following checks have failed and imply this app is pre-patched or obtained from someone else:&lt;br>&lt;br>&lt;small>%1$s&lt;/small>&lt;br>It is strongly recommended to &lt;b>uninstall this app and patch it yourself&lt;/b> to ensure you are using a validated and secure app.&lt;p>&lt;br>&lt;p>If this warning is ignored it will not be shown again.</string>
```

KobeW50: I edited my suggestion a moment before you commited the changes. Instead of my change to `or may be harmful or dangerous` I reverted it back to the original, which was `or could be harmful or dangerous`

Idk. It's very minor, and both ways are correct so whatever

src/main/resources/addresources/values/strings.xml
```diff
@@ -31,6 +31,17 @@ This is because Crowdin requires temporarily flattening this file and removing t
 -->
 <resources>
     <app id="shared">
+        <patch id="misc.checks.BaseCheckEnvironmentPatch">
+            <string name="revanced_check_environment_failed_title">Checks failed</string>
+            <string name="revanced_check_environment_dialog_open_official_source_button">Open official website</string>
+            <string name="revanced_check_environment_dialog_ignore_button">Ignore</string>
+            <string name="revanced_check_environment_failed_message">&lt;h5>This app does not appear to be patched by you.&lt;/h5>&lt;br>This app may not function correctly, &lt;b>could be harmful or even dangerous to use&lt;/b>.&lt;br>&lt;br>The checks have concluded the following and imply this app is pre-patched or obtained from someone else:&lt;br>&lt;br>&lt;small>%1$s&lt;/small>&lt;br>It is strongly recommended to &lt;b>uninstall this app and patch it yourself&lt;/b> to ensure you are using a validated and secure app.&lt;p>&lt;br>&lt;p>If this warning is ignored it will not be shown again.</string>
```
oSumAtrIX, line 38: "will nit be shown again" seems not to be quite correct. The dialog is shown three times until it is not shown anymore.

src/main/kotlin/app/revanced/patches/shared/misc/checks/BaseCheckEnvironmentPatch.kt
```diff
@@ -0,0 +1,116 @@
...
+            Class.forName("android.os.Build")
+            // This only works on Android,
+            // because it uses Android APIs.
+            setBuildInfo()
+        } catch (_: ClassNotFoundException) { }
+    }
+
+    private fun invokeCheck() = mainActivityOnCreateFingerprint.result?.mutableMethod?.addInstructions(
+        0,
+        "invoke-static/range { p0 .. p0 },$INTEGRATIONS_CLASS_DESCRIPTOR->check(Landroid/app/Activity;)V",
```
oSumAtrIX, line 90: Dialogs in integrations can use some queue so they do not overlap with each other. It looks very weird if they do.

src/main/resources/addresources/values/strings.xml
oSumAtrIX: I was more thinking about rewording and telling the user "This warning will only be shown twice if ignored"

oSumAtrIX:
```suggestion
            <string name="revanced_check_environment_failed_message">&lt;h5>This app does not appear to be patched by you.&lt;/h5>&lt;br>This app may not function correctly, &lt;b>could be harmful or even dangerous to use&lt;/b>.&lt;br>&lt;br>The checks have concluded the following and imply this app is pre-patched or obtained from someone else:&lt;br>&lt;br>&lt;small>%1$s&lt;/small>&lt;br>It is strongly recommended to &lt;b>uninstall this app and patch it yourself&lt;/b> to ensure you are using a validated and secure app.&lt;p>&lt;br>&lt;p>This warning will only be shown twice, if ignored.</string>
```

oSumAtrIX:
It is not universal and included for specific apps

PR #3616, opened 2024-09-03
fix(pixiv): `Hide Ads` fingerprint unable to resolve

Anajrim01:
Fixes #3591 

Ads a second fingerprint for newer recent version where the method names aren't obfuscated.

Tested with version 6.121.1 (latest).

Reasoning: A second fingerprint is added to still allow patching of older versions whilst also allowing the patching of newer versions hence maintaining compatibility and if methods are to be obfuscated again then the fingerprints will still work.

### DIFF
--- a/src/main/kotlin/app/revanced/patches/pixiv/ads/HideAdsPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/pixiv/ads/HideAdsPatch.kt
@@ -7,22 +7,28 @@ import app.revanced.patcher.patch.BytecodePatch
 import app.revanced.patcher.patch.annotation.CompatiblePackage
 import app.revanced.patcher.patch.annotation.Patch
 import app.revanced.patches.pixiv.ads.fingerprints.IsNotPremiumFingerprint
+import app.revanced.patches.pixiv.ads.fingerprints.ShouldShowAdsFingerprint
 
 @Patch(
     name = "Hide ads",
     compatiblePackages = [CompatiblePackage("jp.pxv.android")]
 )
 @Suppress("unused")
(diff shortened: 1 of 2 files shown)


oSumAtrIX (changes requested):
I think you can drop support for old versions. Latest is what people are after anyway 👍

Anajrim01:
> I think you can drop support for old versions. Latest is what people are after anyway 👍

Alright, I've removed the old fingerprint and kept the new one only

src/main/kotlin/app/revanced/patches/pixiv/ads/HideAdsPatch.kt
```diff
@@ -6,23 +6,24 @@ import app.revanced.patcher.extensions.InstructionExtensions.addInstructions
...
     name = "Hide ads",
     compatiblePackages = [CompatiblePackage("jp.pxv.android")]
 )
 @Suppress("unused")
-object HideAdsPatch : BytecodePatch(setOf(IsNotPremiumFingerprint)) {
-    // Always return false in the "isNotPremium" method which normally returns !this.accountManager.isPremium.
+object HideAdsPatch : BytecodePatch(setOf(ShouldShowAdsFingerprint)) {
+    // Always return false in the "shouldShowAds" method which normally returns !this.accountManager.isPremium.
     // However, this is not the method that controls the user's premium status.
     // Instead, this method is used to determine whether ads should be shown.
```
oSumAtrIX, line 19: This comment is outdated. shouldShowAds obviously does not control the user's premium status.

Anajrim01: I've removed the comment as the name shouldShowAds reflects what it should do.

PR #3627, opened 2024-09-07
feat(Facebook): Add `Hide sponsored stories` patch

PantlessCoding:
Hello, I created a new patch for Facebook that removes sponsored stories in the main feed.

Tested with com.facebook.katana versions v478.* v480.* any DPI variants. 

This patch will only work with variants Min API 30 (Android 11+) .

TODOs : 
 - Enable hiding of different type of stories/content (ex.: Boosted posts, annoying suggested groups, shorts, etc.)
 - Make this and future patch part of an integration once I figure out what I'm doing

This is my first Android hack and first GitHub contribution ever, so let me know if any comments or remark about this PR as a whole.

Thanks!


PantlessCoding:
@cyberboh Not easily, the problem is that the packages targeting earlier API version have an almost completely different code base. The classes getting patched here simply does not exist.  It's the same issue with the other Facebook patch you reference. 
The newer android API versions seems like a complete native code rewrite versus the use of a web container. Might be doable, but with a different approach that require a lot more investigation.

src/main/kotlin/app/revanced/patches/facebook/ads/mainfeed/HideSponsoredStoriesPatch.kt
```diff
@@ -0,0 +1,42 @@
...
+import app.revanced.patcher.extensions.InstructionExtensions.addInstructionsWithLabels
+import app.revanced.patcher.patch.BytecodePatch
+import app.revanced.patcher.patch.annotation.CompatiblePackage
+import app.revanced.patcher.patch.annotation.Patch
+import app.revanced.patches.facebook.ads.mainfeed.fingerprints.GetStoryVisibilityFingerprint
+import app.revanced.patches.facebook.ads.mainfeed.fingerprints.GraphQLStorySponsoredDataGetterFingerprint
+
+@Patch(
+    name = "Hide Sponsored stories",
+    description = "Hide Sponsored stories in main feed.",
```
oSumAtrIX, line 14:
```suggestion
```

src/main/kotlin/app/revanced/patches/facebook/ads/mainfeed/HideSponsoredStoriesPatch.kt
```diff
@@ -0,0 +1,42 @@
...
+    compatiblePackages = [CompatiblePackage("com.facebook.katana")]
+)
+@Suppress("unused")
+object HideSponsoredStoriesPatch : BytecodePatch(setOf(GetStoryVisibilityFingerprint, GraphQLStorySponsoredDataGetterFingerprint)) {
+    override fun execute(context: BytecodeContext) {
+        GetStoryVisibilityFingerprint.result?.apply {
+
+            // Get the sponsored data model getter
+            val sponsoredDataModelGetter = GraphQLStorySponsoredDataGetterFingerprint.result ?: throw GraphQLStorySponsoredDataGetterFingerprint.exception;
```
oSumAtrIX, line 23: Don't forget to import

```suggestion
            val sponsoredDataModelGetter = GraphQLStorySponsoredDataGetterFingerprint.resultOrThrow();
```

src/main/kotlin/app/revanced/patches/facebook/ads/mainfeed/HideSponsoredStoriesPatch.kt
```diff
@@ -0,0 +1,42 @@
...
+    name = "Hide Sponsored stories",
+    description = "Hide Sponsored stories in main feed.",
+    compatiblePackages = [CompatiblePackage("com.facebook.katana")]
+)
+@Suppress("unused")
+object HideSponsoredStoriesPatch : BytecodePatch(setOf(GetStoryVisibilityFingerprint, GraphQLStorySponsoredDataGetterFingerprint)) {
+    override fun execute(context: BytecodeContext) {
+        GetStoryVisibilityFingerprint.result?.apply {
+
+            // Get the sponsored data model getter
```
oSumAtrIX, line 22: Comments end with punctuation
```suggestion
            // Get the sponsored data model getter.
```

src/main/kotlin/app/revanced/patches/facebook/ads/mainfeed/HideSponsoredStoriesPatch.kt
```diff
@@ -0,0 +1,42 @@
...
+            val sponsoredDataModelGetter = GraphQLStorySponsoredDataGetterFingerprint.result ?: throw GraphQLStorySponsoredDataGetterFingerprint.exception;
+
+            // Hide stories with sponsored data defined
+            // Check if param type is GraphQLStory
+            // If so, calling the sponsoredDataModelGetter, only sponsored content has non-null data
+            // Sponsored Stories gets their visibility forced to StoryVisibility.GONE
```
oSumAtrIX, line 28:
```suggestion

            // Check if the parameter type is GraphQLStory and if sponsoredDataModelGetter returns a non-null value.
            // If so, hide the story by setting the visibility to StoryVisibility.GONE.
```

src/main/kotlin/app/revanced/patches/facebook/ads/mainfeed/HideSponsoredStoriesPatch.kt
```diff
@@ -0,0 +1,42 @@
...
+                    if-eqz v0, :resume_normal
+                    invoke-virtual {p0}, Lcom/facebook/graphql/model/GraphQLStory;->${sponsoredDataModelGetter.method.name}()${sponsoredDataModelGetter.method.returnType}
+                    move-result-object v0 
+                    if-eqz v0, :resume_normal
+                    const-string v0, "GONE"
+                    return-object v0
+                    :resume_normal
+                    nop
+            """)
+        } ?: throw GetStoryVisibilityFingerprint.exception
```
oSumAtrIX, line 40:
```suggestion
            mutableMethod.addInstructionsWithLabels(
	            scanResult.patternScanResult!!.startIndex,
	            """
	            	instance-of v0, p0, Lcom/facebook/graphql/model/GraphQLStory;
	                if-eqz v0, :resume_normal
	                invoke-virtual {p0}, Lcom/facebook/graphql/model/GraphQLStory;->${sponsoredDataModelGetter.method.name}()${sponsoredDataModelGetter.method.returnType}
	                move-result-object v0 
	                if-eqz v0, :resume_normal
	                const-string v0, "GONE"
	                return-object v0
	                :resume_normal
	                nop
	            """
	        )
        } ?: throw GetStoryVisibilityFingerprint.exception
```

src/main/kotlin/app/revanced/patches/facebook/ads/mainfeed/fingerprints/GraphQLStorySponsoredDataGetterFingerprint.kt
```diff
@@ -0,0 +1,18 @@
...
+import app.revanced.patcher.fingerprint.MethodFingerprint
+import com.android.tools.smali.dexlib2.AccessFlags
+import com.android.tools.smali.dexlib2.Opcode
+import com.android.tools.smali.dexlib2.iface.Annotation
+import com.android.tools.smali.dexlib2.iface.instruction.formats.Instruction31i
+import com.android.tools.smali.dexlib2.iface.value.StringEncodedValue
+
+internal object GraphQLStorySponsoredDataGetterFingerprint : MethodFingerprint(
+    customFingerprint = { methodDef, classDef ->
+        // All methods within initialise a generic model with the same instructions apart from the value of those two constants
```
oSumAtrIX, line 13: The comment is not clear. Please check if you can reword.
```suggestion
        // All methods within initialize a generic model with the same instructions apart from the value of those two constants
```

src/main/kotlin/app/revanced/patches/facebook/ads/mainfeed/fingerprints/GraphQLStorySponsoredDataGetterFingerprint.kt
```diff
@@ -0,0 +1,18 @@
...
+        classDef.type == "Lcom/facebook/graphql/model/GraphQLStory;" &&
+                (methodDef.implementation?.instructions?.elementAt(1) as? Instruction31i)?.narrowLiteral == -132939024 &&
+                (methodDef.implementation?.instructions?.elementAt(2) as? Instruction31i)?.narrowLiteral == 341202575
```
oSumAtrIX, line 16: Reduce duplication via a variable

src/main/kotlin/app/revanced/patches/facebook/ads/mainfeed/fingerprints/GraphQLStorySponsoredDataGetterFingerprint.kt
```diff
@@ -0,0 +1,18 @@
...
+import com.android.tools.smali.dexlib2.Opcode
+import com.android.tools.smali.dexlib2.iface.Annotation
+import com.android.tools.smali.dexlib2.iface.instruction.formats.Instruction31i
+import com.android.tools.smali.dexlib2.iface.value.StringEncodedValue
+
+internal object GraphQLStorySponsoredDataGetterFingerprint : MethodFingerprint(
+    customFingerprint = { methodDef, classDef ->
+        // All methods within initialise a generic model with the same instructions apart from the value of those two constants
+        classDef.type == "Lcom/facebook/graphql/model/GraphQLStory;" &&
+                (methodDef.implementation?.instructions?.elementAt(1) as? Instruction31i)?.narrowLiteral == -132939024 &&
```
oSumAtrIX, line 15: If possible name those magic constants. What are those?

PantlessCoding: I have done some more digging to better document this. They appear to be GraphQL type ids. I am pretty sure I have the correct name for "341202575" as there is a lookup for the string "sponsored_data" that resolves to it. The other one is still speculative, but this is what makes the most sense to me at the moment. I also updated the comments explaining those findings.

src/main/kotlin/app/revanced/patches/facebook/ads/mainfeed/HideSponsoredStoriesPatch.kt
```diff
@@ -0,0 +1,43 @@
...
+            // Get the sponsored data model getter.
+            val sponsoredDataModelGetter = GraphQLStorySponsoredDataGetterFingerprint.resultOrThrow()
+
+            // Check if the parameter type is GraphQLStory and if sponsoredDataModelGetter returns a non-null value.
+            // If so, hide the story by setting the visibility to StoryVisibility.GONE.
+            mutableMethod.addInstructionsWithLabels(
+	            scanResult.patternScanResult!!.startIndex,
+	            """
+	            	instance-of v0, p0, Lcom/facebook/graphql/model/GraphQLStory;
+	                if-eqz v0, :resume_normal
```
oSumAtrIX, line 31: There's inconsistent use of tabs and spaces here, please use spaces

src/main/kotlin/app/revanced/patches/facebook/ads/mainfeed/HideSponsoredStoriesPatch.kt
```diff
@@ -0,0 +1,43 @@
...
+import app.revanced.patcher.extensions.InstructionExtensions.addInstructionsWithLabels
+import app.revanced.patcher.patch.BytecodePatch
+import app.revanced.patcher.patch.annotation.CompatiblePackage
+import app.revanced.patcher.patch.annotation.Patch
+import app.revanced.patches.facebook.ads.mainfeed.fingerprints.GetStoryVisibilityFingerprint
+import app.revanced.patches.facebook.ads.mainfeed.fingerprints.GraphQLStorySponsoredDataGetterFingerprint
+import app.revanced.util.resultOrThrow
+
+@Patch(
+    name = "Hide Sponsored stories",
```
oSumAtrIX, line 14:
```suggestion
    name = "Hide Sponsored Stories",
```

oSumAtrIX:
I've refactored the code. Please test the changes. Apart from that LGTM.

src/main/kotlin/app/revanced/patches/facebook/ads/mainfeed/fingerprints/GraphQLStorySponsoredDataGetterFingerprint.kt
```diff
@@ -0,0 +1,29 @@
...
+        // All methods within the target class are virtually identical apart from two magic constants on each of them.
+        // They all return an instance of the same model but filled with different GraphQL child node data.
+        // Likely, the constants are the parent and child node ids.
+        classDef.type == "Lcom/facebook/graphql/model/GraphQLStory;" &&
+                isMatchingLiteralValue(1, STORY_TYPE_ID) &&
+                isMatchingLiteralValue(2, SPONSORED_DATA_TYPE_ID)
+    }
+) {
+    // These constants trace back to some GraphQL node type descriptor.
+    private const val STORY_TYPE_ID = -132939024
```
oSumAtrIX, line 27: Are these the same in older versions? If these values change, we can't use them.

PantlessCoding: Sorry for the delay, all the code revision still work. I checked the latest Facebook version and the last 3 major, the ids are the same. But, I agree that this is not ideal and I'd like to make this resolve dynamically. I might have found a way, I will try to implement it shortly.

PantlessCoding:
Made it, it became complicated quickly! Basically I found another method that retrieve SponsoredData with the model ids I need but for a different sub-type of entity.  The method is easy to fingerprint, but I can't use it directly since I can't cast to the correct type in my context. Instead I just grab the ids and call the base method directly via a helper method. The helper also cast the sponsored data to the generic type, this is fine since we only check for null, but it makes it so it should work on any entity that has sponsored data. Man that app is heavily obfuscated...

oSumAtrIX (approved):
Would it be possible to move the method you are adding to the class to integrations? It may not add much value though since you need to patch that method anyways to insert the instructions, so I'll merge as is.

PantlessCoding:
> Would it be possible to move the method you are adding to the class to integrations? It may not add much value though since you need to patch that method anyways to insert the instructions, so I'll merge as is.

No problem, will see what can be done on future patches. Next, I really would like to find a way to support API29 variants. On that note, is there a way to specify minimum API version on patches like for the version number ? If not would it be relevant to add ? 
Also how should I handle patches for a an app that has a different code base on some variants, That would be based on if some fingerprints resolves or not ?
Thanks !

oSumAtrIX:
A patch can read the minsdk from the manifest. But why would the patch not work on versions below that. 

If two apps have different code base, you can abstract your code so it fits both. You can slso use share a fingerprint between pathes

PantlessCoding:
Thanks. In this case, Facebook release simultaneously  a minAPI29 and a minAPI30 version. From my initial look, it seems like the main reason patches does not work on API29 (the other Facebook patch included) is that the classes are "superpacked" and decompressed at runtime. After that, it is likely that it's the same code in the end. I will see what can be done with that.

oSumAtrIX:
That sounds reasonable. A patch can technically decompress the dex, modify them and then compress them back, but this would need some boilerplate work which isn't really worth it.

Ollibolli85:
Sadly, some points in menu do not work, if i install (not mount) the patched app!

There are also sponsored links in the patched app!

Both on Samsung, Android 14!

mindplay-dk:
> This patch will only work with variants Min API 30 (Android 11+) .

for the record, it's listed as "works with any version" - I came here trying to find out why it failed to patch.

you should probably add a note explaining which versions to use it with.

mindplay-dk:
hmm, it failed with the latest Android 11+ release as well. maybe it used to work? or else I don't know what I'm doing wrong - there is an endless list of releases, I think you need to specify what this is expected to work work. I couldn't find a recent release that works.

Issue #3619, opened 2024-09-04
bug: Soundcloud patch "Hide ADS" no longer works


alan7383:
### Bug description

From Soundcloud version 2024.09.03-beta the "Hide ADS" patch no longer works, see logs.

### Error logs

```shell
- Device Info
ReVanced Manager: 1.21.2
Model: 22101316G
Android version: 14
Supported architectures: arm64-v8a, armeabi-v7a, armeabi
Root permissions: Yes

- Patch Info
App: com.soundcloud.android v2024.09.03-beta (Suggested: Any)
Patches version: v4.13.3
Patches added: Default
Patches removed: None
Default patch options changed: None

- Settings
Allow changing patch selection: false
Version compatibility check: true
Show universal patches: false
Patches source: revanced/revanced-patches
Integration source: revanced/revanced-integrations

- Logs
Reading APK
Decoding app manifest
Loading patches
Executing patches
Applied 2 patches
Hide ads failed: app.revanced.patcher.patch.PatchException: Failed to resolve InterceptFingerprint
at app.revanced.util.BytecodeUtilsKt.getException(BytecodeUtils.kt:27)
at app.revanced.util.BytecodeUtilsKt.resultOrThrow(BytecodeUtils.kt:19)
at app.revanced.patches.soundcloud.ad.HideAdsPatch.execute(HideAdsPatch.kt:64)
at app.revanced.patches.soundcloud.ad.HideAdsPatch.execute(HideAdsPatch.kt:17)
at app.revanced.patcher.Patcher$apply$1.invokeSuspend$executePatch(Unknown Source:190)
at app.revanced.patcher.Patcher$apply$1.invokeSuspend(Unknown Source:282)
at app.revanced.patcher.Patcher$apply$1.invoke(SourceFile:0)
at app.revanced.patcher.Patcher$apply$1.invoke(SourceFile:0)
at kotlinx.coroutines.flow.SafeFlow.collectSafely(Unknown Source:2)
at kotlinx.coroutines.flow.AbstractFlow.collect(Unknown Source:71)
at app.revanced.manager.flutter.MainActivity$runPatcher$1$patcherResult$1$2.invokeSuspend(Unknown Source:76)
at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(Unknown Source:11)
at kotlinx.coroutines.DispatchedTask.run(Unknown Source:98)
[...]

PR #3628, opened 2024-09-07
fix(Soundcloud - Hide ads): Support latest version

FullerBread2032:
- InterceptFingerprint is now resolving in newer versions (#3619)

Minimum version: `2024.09.03-beta`

Unfortunately, this pull request currently only works with the latest beta versions and makes older versions stop working. Is there an option to add an old fingerprint to support older versions? If so, how would I do that?


BenCat07:
... i seem to have missed this Pull request. Woops. Closing #3642

oSumAtrIX (approved):

PR #3637, opened 2024-09-13
feat(YouTube): Unlock sleep timer menu

zainarbani:
<details>
<summary>Screenshots</summary>
 
image1

image2

</details>

### DIFF
--- a/src/main/kotlin/app/revanced/patches/youtube/layout/buttons/player/flyout/SleepTimerMenuPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/youtube/layout/buttons/player/flyout/SleepTimerMenuPatch.kt
@@ -1,0 +1,49 @@
+package app.revanced.patches.youtube.layout.buttons.player.flyout
+
+import app.revanced.patcher.data.BytecodeContext
+import app.revanced.patcher.extensions.InstructionExtensions.addInstruction
+import app.revanced.patcher.extensions.InstructionExtensions.addInstructions
+import app.revanced.patcher.patch.BytecodePatch
+import app.revanced.patcher.patch.annotation.CompatiblePackage
+import app.revanced.patcher.patch.annotation.Patch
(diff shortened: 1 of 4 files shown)


zainarbani:
> If you press `Close`, will YouTube close, and will you no longer be able to open it?

If the user still 'awake' then press close to continue watching, or simply ignore it (sleep) then screen will go off by itself.

src/main/kotlin/app/revanced/patches/youtube/layout/buttons/player/flyout/SleepTimerMenuPatch.kt
```diff
@@ -0,0 +1,49 @@
...
+import app.revanced.util.resultOrThrow
+
+@Patch(
+    name = "Remove sleep timer menu restrictions",
+    description = "Show sleep timer options in player flyout menu.",
+    compatiblePackages = [
+        CompatiblePackage(
+            "com.google.android.youtube",
+            [
+                "19.30.39", // Only enable this patch on 19.30+ for now.
```
oSumAtrIX, line 20: Why?

zainarbani:
I just realized that this features was now enabled by default on stock 19.41.39, also on non premium (tested with 2 account).
Can you reproduce this? If this applied to all user then we don't need this patches i think.

zainarbani:
@LisoUseInAIKyrios @MarcaDian I think duplicated menu is fixed now, can you confirm on your end?

zainarbani:
https://9to5google.com/2024/10/15/youtube-new-miniplayer/

zainarbani:
@MarcaDian Try clearing stock app data.

zainarbani:
Off topic but it seems modern miniplayer & new cairo setting layout are now enabled by default on stock app.

zainarbani:
I assume its pretty safe to remove the feature flag patch now

zainarbani:
@oSumAtrIX @LisoUseInAIKyrios btw, is there a plan to enable the new cairo settings layout? just in case the old layout removed by youtube.

PR #3642, opened 2024-09-14
Fix(Soundcloud): 'Hide Ads' Patch

BenCat07:
This should work the same... but unfortunately soundcloud now uses split apks, making this near impossible to check for me.

Proper Test is still needed.


BenCat07:
#3628 Already implements this.

Issue #3696, opened 2024-09-26
bug(Soudcloud - Disable telemetry): Patching fails


szbinks:
### Bug description

I tried to patch the last version of Soundcloud (2024.09.20) but the patch "Disable Telemetry failed.

To do so I have downloaded Soundcloud on apkmirror, then I use AntiSplit M to convert the apkm file to an apk. Finaly I used Revanced Manager on Android to patch the apk file.

### Error logs

```shell
- Device Info
ReVanced Manager: 1.22.0
Model: SM-G991B
Android version: 14
Supported architectures: arm64-v8a, armeabi-v7a, armeabi
Root permissions: No

- Patch Info
App: com.soundcloud.android v2024.09.20-release (Suggested: Any)
Patches version: v4.15.0
Patches added: Default
Patches removed: None
Default patch options changed: None

- Settings
Allow changing patch selection: true
Version compatibility check: true
Show universal patches: false
Patches source: revanced/revanced-patches
Integration source: revanced/revanced-integrations

- Logs
Reading APK
Decoding app manifest
Loading patches
Executing patches
Disable telemetry failed: app.revanced.patcher.patch.PatchException: Failed to resolve CreateTrackingApiFingerprint
	at app.revanced.util.BytecodeUtilsKt.getException(BytecodeUtils.kt:27)
	at app.revanced.util.BytecodeUtilsKt.resultOrThrow(BytecodeUtils.kt:19)
	at app.revanced.patches.soundcloud.analytics.DisableTelemetryPatch.execute(DisableTelemetryPatch.kt:22)
	at app.revanced.patches.soundcloud.analytics.DisableTelemetryPatch.execute(DisableTelemetryPatch.kt:11)
	at app.revanced.patcher.Patcher$apply$1.invokeSuspend$executePatch(Unknown Source:190)
	at app.revanced.patcher.Patcher$apply$1.invokeSuspend(Unknown Source:282)
	at app.revanced.patcher.Patcher$apply$1.invoke(SourceFile:0)
	at app.revanced.patcher.Patcher$apply$1.invoke(SourceFile:0)
	at kotlinx.coroutines.flow.SafeFlow.collectSafely(Unknown Source:2)
	at kotlinx.coroutines.flow.AbstractFlow.collect(Unknown Source:71)
	at app.revanced.manager.flutter.MainActivity$runPatcher$1$patcherResult$1$2.invokeSuspend(Unknown Source:76)
[...]

PR #3702, opened 2024-09-27
fix(Soundcloud): Support version `2024.09.20`

FullerBread2032:
- Fingerprints are now resolved in the latest version ( #3696 )

TODO: 
- [x] Check if the patches are working
- [x] Compatibility with older versions

Tested versions: `2024.09.12`, `2024.09.16`, `2024.09.20`


oSumAtrIX:
I'd say we can do this the next time it breaks. The fingerprints were relying on names so without those they should unlikely break like this again.

oSumAtrIX (approved):

PR #3752, opened 2024-10-08
feat(YouTube - Hide Shorts components): Hide `Use template`, `Upcoming`, `Green screen` buttons

MarcaDian:
Closes https://github.com/ReVanced/revanced-patches/issues/3751

[Integration changes](https://github.com/ReVanced/revanced-integrations/pull/714)

### DIFF
--- a/src/main/resources/addresources/values/strings.xml
+++ b/src/main/resources/addresources/values/strings.xml
@@ -639,6 +639,9 @@ This is because Crowdin requires temporarily flattening this file and removing t
             <string name="revanced_hide_shorts_save_sound_button_title">Hide save music button</string>
             <string name="revanced_hide_shorts_save_sound_button_summary_on">Save music is hidden</string>
             <string name="revanced_hide_shorts_save_sound_button_summary_off">Save music is shown</string>
+            <string name="revanced_hide_shorts_use_template_button_title">Hide use template button</string>
+            <string name="revanced_hide_shorts_use_template_button_summary_on">Use template is hidden</string>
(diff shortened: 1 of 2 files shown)


src/main/resources/addresources/values/strings.xml
```diff
@@ -639,6 +639,9 @@ This is because Crowdin requires temporarily flattening this file and removing t
             <string name="revanced_hide_shorts_save_sound_button_title">Hide save music button</string>
             <string name="revanced_hide_shorts_save_sound_button_summary_on">Save music is hidden</string>
             <string name="revanced_hide_shorts_save_sound_button_summary_off">Save music is shown</string>
+            <string name="revanced_hide_shorts_use_template_button_title">Hide use template button</string>
+            <string name="revanced_hide_shorts_use_template_button_summary_on">Use template is hidden</string>
```
ILoveOpenSourceApplications, line 643:
```suggestion
            <string name="revanced_hide_shorts_use_template_button_summary_on">Use template button is hidden</string>
```
Also, unrelated to this PR, but the same change is needed for `Save music` button as well.

ILoveOpenSourceApplications: But if https://github.com/ReVanced/revanced-patches/issues/3751#issuecomment-2400226068 is implemented then these changes won't be necessary.

ILoveOpenSourceApplications: The summary part is the one which requires changes. The button title is fine.

PR #3804, opened 2024-10-23
feat(YouTube - Hide layout components): Hide player shopping shelf

LisoUseInAIKyrios:
With 19.24 and later, the video player can show under the video a shopping shelf with merchandise from the channel.

Enabling `hide horizontal shelves` will hide this, but this can be selectively hidden since this is the only shelf that can appear in the video player.

This PR adds a new hide ads setting to hide this specific type of shelf (enabled by default).

shopping shelf

[Integration changes](https://github.com/ReVanced/revanced-integrations/pull/723)


oSumAtrIX (approved):
Probably should be a feat.

PR #3815, opened 2024-10-25
feat(Discord): Add `Plugin loader` patch

pjanssenrammsteinn:
Debug true


src/main/kotlin/app/revanced/patches/shared/misc/react/fingerprints/LoadScriptFromAssetsFingerprint.kt
```diff
@@ -0,0 +1,3 @@
+package app.revanced.patches.shared.misc.react.fingerprints
```
pjanssenrammsteinn, line 1: Debug

PR #3819, opened 2024-10-26
ci: Use updated paths for translating patches 

oSumAtrIX:


oSumAtrIX:
I don't think it's needed. Pulling can be done by dispatching the workflow manually, right before the dev branch is merged.

Issue #2545, opened 2023-12-27
bug(Change package name): Changing package name causing collisions within it's own manifest


Shadow-Hunt3r:
### Bug description

When applying the patch "Custom Package Name" to certain apps you can be faced with one of 2 issues depending on the app
AndroidManifest containing providers such as `<provider android:name="com.facebook.FacebookContentProvider" ... />` which could be a collision to the app while still running the original package name throughout the manifest

Steps to Reproduce Provider Collision:
1. Install and extract the base apk for an app with a provider name as above
2. Apply patch change package name modification to the base apk
3. Installing would cause a collision, in games it could occur regardless if the original game is still installed or not

After removing the provider above with the first solution below I closed the apk and ran it through revanced manager again to find another issue of "Can't install because provider name com.original.apk.NativeShareContentProvider (in package com.modded.apk) is already used by com.original.apk"
This issue is most likely caused by lines in the Manifest that follow as such `android:authorities="com.original.apk.PERMISSION"` when the permissions are given to com.original.apk

### Solution

For the ContentProvider collision, i believe it to be caused as it is usually supplied with an identifier given with the app, making a patch that can find and remove or modify the code below could fix this issue, while warning that the feature may not work anymore based on how it is completed
 ```
  <provider android:name="com.facebook.FacebookContentProvider"
              android:exported="true"
              android:authorities="com.facebook.app.FacebookContentProvider*ID*" />
```
For changing the ID there would also be a meta-data string that contains the ID as well under the following
```
     <meta-data android:name="com.facebook.sdk.ApplicationId"
               android:value="fb*ID*" />
```

For the second issue parsing the manifest and replacing all traces of `android:authorities="com.original.apk.PERMISSION"` and replacing it with `android:authorities="com.modded.apk.PERMISSION"` seems to have fixed the issue but at the cost of Google Services in the app to no longer work, however it could possibly be fixed by adding seeing if it is possible to make GmsCore Support to have a universal application

### Additional context

Below I have shown such an example using a game called Hackers by Trickster Arts
Before Removing FB - Fail caused by first error provided above

```
- Device Info
ReVanced Manager: 1.18.0
Model: Pixel 6a
Android version: 14
Supported architectures: arm64-v8a, armeabi-v7a, armeabi
Root permissions: No

- Patch Info
App: com.tricksterarts.hackers v1.225 (Suggested: Any)
Patches version: v3.1.0
[...]

PR #3846, opened 2024-10-31
feat(Change package name): Change provider package names

1fexd:
The existing `Change package name` patch only renames the package attribute of the manifest. If encountered some problems while trying to use it to clone the Mega.nz app, due to its use of providers and the `DYNAMIC_RECEIVER_NOT_EXPORTED_PERMISSION` permission, so I extended the patch to also update those. 

Cloning Mega seems to work now, although I'm not entirely sure if there are any problems at runtime due to the renamed provider authorities


1fexd:
> Would this also fix issue #2545?

Just tested Facebook 449 but it doesn't even decode its resources:

```
INFO: Decoding resources 
java.lang.IllegalArgumentException: expected element name 'view' and not 'View'
	at org.xmlpull.renamed.MXSerializer.endTag(MXSerializer.java:708)
```

oSumAtrIX:
PS: Decoding is handled by Androlib by ApkTool. So this issue should be mentioned there.

1fexd:
> It appears this change can break embedded usage of other apps, such as Facebook integration within an app.

In which app have you observed this behavior? Maybe that's fixable

> This change should be a patch option that is default off. The option should explain that enabling it can fix installation issues if the patched app fails to install, but the option can break some in app features.

Agreed

oSumAtrIX (approved):

oSumAtrIX (approved):

PR #3869, opened 2024-11-06
chore(YouTube - Announcements): Use API v4 to get announcements

oSumAtrIX:
This PR updates the Announcements patch to use API v4. With v4 now, the ID can be requested only. Announcements also now have an archived_at field. If the announcement is archived, even if it's the latest and has not been read, it will not be shown. This is useful for new users when the announcement is no longer relevant.

### DIFF
--- a/extensions/shared/src/main/java/app/revanced/extension/youtube/patches/announcements/AnnouncementsPatch.java
+++ b/extensions/shared/src/main/java/app/revanced/extension/youtube/patches/announcements/AnnouncementsPatch.java
@@ -2,7 +2,8 @@ package app.revanced.extension.youtube.patches.announcements;
 
 import static android.text.Html.FROM_HTML_MODE_COMPACT;
 import static app.revanced.extension.shared.StringRef.str;
-import static app.revanced.extension.youtube.patches.announcements.requests.AnnouncementsRoutes.GET_LATEST_ANNOUNCEMENT;
+import static app.revanced.extension.youtube.patches.announcements.requests.AnnouncementsRoutes.GET_LATEST_ANNOUNCEMENTS;
+import static app.revanced.extension.youtube.patches.announcements.requests.AnnouncementsRoutes.GET_LATEST_ANNOUNCEMENT_IDS;
 
 import android.app.Activity;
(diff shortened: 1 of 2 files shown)


extensions/shared/src/main/java/app/revanced/extension/youtube/patches/announcements/AnnouncementsPatch.java
```diff
@@ -39,51 +82,45 @@ public static void showAnnouncement(final Activity context) {
...
                     message = jsonString;
                 }
 
-                // Do not show the announcement, if the last announcement id is the same as the current one.
-                if (Settings.ANNOUNCEMENT_LAST_ID.get() == id) return;
+                // If the announcement is archived, do not show it.
+                if (archivedAt.isBefore(LocalDateTime.now())) {
+                    Settings.ANNOUNCEMENT_LAST_ID.save(id);
+                    return;
+                }
```
oSumAtrIX, line 123: Mark the announcement as read if archived.

extensions/shared/src/main/java/app/revanced/extension/youtube/patches/announcements/requests/AnnouncementsRoutes.java
```diff
@@ -9,12 +9,9 @@
...
 
 public class AnnouncementsRoutes {
-    private static final String ANNOUNCEMENTS_PROVIDER = "https://api.revanced.app/v2";
-
-    /**
-     * 'language' parameter is IETF format (for USA it would be 'en-us').
-     */
-    public static final Route GET_LATEST_ANNOUNCEMENT = new Route(GET, "/announcements/youtube/latest?language={language}");
+    public static final Route GET_LATEST_ANNOUNCEMENTS = new Route(GET, "/announcements/latest?tag=youtube");
+    public static final Route GET_LATEST_ANNOUNCEMENT_IDS = new Route(GET, "/announcements/latest/id?tag=youtube");
```
oSumAtrIX, line 13: An ID route is used to not get the entire announcement every time to check if it's the latest.

PR #3905, opened 2024-11-12
chore: Separate extensions by app

oSumAtrIX:
# Todo

- [ ] Adjust ProGuard files, build.gradle.kts and AndroidManifest of each extension
- [ ] Fix modular design (shared project depends on app-specific projects for example when it should not)


oSumAtrIX:
The shared library still contains shared code specific to two Reddit apps for example, which could be moved out, but this can be done in a separate PR:

image

Apps now can use their own extension which also merges the shared extension for the necessary APIs:

image

oSumAtrIX:
Looks like one modularity issue was added recently:

https://github.com/ReVanced/revanced-patches/actions/runs/12037946631/job/33562511520?pr=3905#step:5:111

PR #3990, opened 2024-11-26
fix(YouTube - Playback speed): Allow long press 2x speed when using custom playback speeds

LisoUseInAIKyrios:
Fixes tap and hold for 2x speed not working when using a default playback speed.

If using custom playback speeds and the current speed is higher than 2.0x, then the hold for 2x effectively does nothing.  It's caused by the logic in YouTube assuming the playback speed cannot be above it's unpatched max of 2.0x

### DIFF
--- a/extensions/shared/src/main/java/app/revanced/extension/youtube/patches/playback/speed/RememberPlaybackSpeedPatch.java
+++ b/extensions/shared/src/main/java/app/revanced/extension/youtube/patches/playback/speed/RememberPlaybackSpeedPatch.java
@@ -20,12 +20,8 @@ public final class RememberPlaybackSpeedPatch {
      * Injection point.
      */
     public static void newVideoStarted(VideoInformation.PlaybackController ignoredPlayerController) {
-        try {
-            Logger.printDebug(() -> "newVideoStarted");
-            newVideoStarted = true;
-        } catch (Exception ex) {
-            Logger.printException(() -> "newVideoStarted failure", ex);
-        }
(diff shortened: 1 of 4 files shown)


extensions/shared/src/main/java/app/revanced/extension/youtube/patches/playback/speed/RememberPlaybackSpeedPatch.java
```diff
@@ -29,42 +31,56 @@ public static void newVideoStarted(VideoInformation.PlaybackController ignoredPl
...
-                if (Settings.PLAYBACK_SPEED_DEFAULT.get() == finalPlaybackSpeed) {
-                    // User changed to a different speed and immediately changed back.
-                    // Or the user is going past 8.0x in the glitched out 0.05x menu.
-                    return;
-                }
-                Settings.PLAYBACK_SPEED_DEFAULT.save(finalPlaybackSpeed);
-
-                Utils.showToastLong(str("revanced_remember_playback_speed_toast", (finalPlaybackSpeed + "x")));
-            }, TOAST_DELAY_MILLISECONDS);
+        try {
```
oSumAtrIX, line 34: The try block doesn't seem necessary here

oSumAtrIX:
I have set the speed to 2.1, but on long press the playback speed still stays 2.1.

oSumAtrIX:
What does this PR specifically solve, though? I assumed it would fix long press 2x speed on custom speeds higher than 2x based on the PR description.

oSumAtrIX (approved):

PR #4000, opened 2024-11-27
fix(YouTube - Spoof video streams): Log out the iOS client to allow kids videos playback

oSumAtrIX:
YouTube now has restricted the auth token to their respective client. This mean's the Android token can not be used for iOS or VR requests anymore (More info https://github.com/yt-dlp/yt-dlp/issues/11640#issuecomment-2501049515).

Private or kids' videos can not be played without logging in.

The only solution here is to ask the user to log in, or somehow handle the login to other clients when the user logs into GmsCore.

Currently, VR seems to work with the Android auth token. This means, that private videos will work. What won't work is private kids videos, because VR is the only authorized client currently,  and that one can't play kids videos.

### DIFF
--- a/extensions/shared/src/main/java/app/revanced/extension/youtube/patches/spoof/ClientType.java
+++ b/extensions/shared/src/main/java/app/revanced/extension/youtube/patches/spoof/ClientType.java
@@ -9,7 +9,7 @@ import androidx.annotation.Nullable;
 
 public enum ClientType {
     // Public videos, but no kids, private or age restricted videos.
-    ANDROID_VR_UNAUTHORIZED(28,
+    ANDROID_VR(28,
             "Quest 3",
             "12",
             "com.google.android.apps.youtube.vr.oculus/1.56.21 (Linux; U; Android 12; GB) gzip",
@@ -19,7 +19,7 @@ public enum ClientType {
             false
     ),
     // Specific purpose for age restricted videos.
-    ANDROID_VR(28,
+    ANDROID_VR_LOGGED_IN(28,
             "Quest 3",
             "12",
             "com.google.android.apps.youtube.vr.oculus/1.56.21 (Linux; U; Android 12; GB) gzip",
@@ -92,7 +92,7 @@ public enum ClientType {
     /**
      * If the client can access the API logged in.
      */
-    public final boolean canAuthorize;
+    public final boolean canLogin;
 
     ClientType(int id,
                String deviceModel,
@@ -101,7 +101,7 @@ public enum ClientType {
                @Nullable String androidSdkVersion,
                String clientVersion,
                String clientName,
-               boolean canAuthorize
+               boolean canLogin
     ) {
         this.id = id;
         this.deviceModel = deviceModel;
@@ -110,6 +110,6 @@ public enum ClientType {
         this.androidSdkVersion = androidSdkVersion;
         this.clientVersion = clientVersion;
         this.clientName = clientName;
-        this.canAuthorize = canAuthorize;
+        this.canLogin = canLogin;
     }
 }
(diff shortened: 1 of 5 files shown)


oSumAtrIX:
@LisoUseInAIKyrios If the first requests fail, should a toast be shown, even if the second or third attempt worked?

oSumAtrIX:
The only way I am seeing to restore playback on private or restricted videos is to use a client that can authorize using the Android Bearer token. Since iOS and VR don't accept the auth token from Android, the only way to play those videos with other clients is to ask the user to login to get the auth token via the auth method used for the other clients. This is not really elegant since the user would have to login twice, once for the Android app, and once for the private videos. The alternative solution is to generate a login for other clients via GmsCore. This would require implementing additional code when the user logs into the app, so that it also manages the login of the user for the other clients.

Karthik-Subbu:
Should Android TV client be added to support playback of kids videos (private/public) in addition to IOS logged out?

oSumAtrIX:
Android VR can not play kids videos. Public or private. Only iOS can. And since iOS is not authorized, private kids video can not be played on any circumstance. This is a limitation unless we can somehow make an authorized request with a client other than Android VR but since the auth token is constrained to it's platform, that's not gonna be easy

Karthik-Subbu:
Could android tv client with existing auth token play private kids video?

oSumAtrIX:
If they restrict auth to their respective platform, probably not.

kitadai31:
> Should Android TV client be added to support playback of kids videos (private/public) in addition to IOS logged out?

I also thought the same thing

In ReVanced Extended, kids videos is playing with Android TV client correctly
It tries in the order of iOS -> Android VR -> Android TV

```
D/Extended: StreamingDataRequest: Fetching video streams for: 5gZOYKHXwyQ using client: IOS
I/Extended: StreamingDataRequest: IOS not available with response code: 400 message: Bad Request
D/Extended: StreamingDataRequest: video: 5gZOYKHXwyQ took: 123ms
D/Extended: StreamingDataRequest: Fetching video streams for: 5gZOYKHXwyQ using client: ANDROID_VR
D/Extended: StreamingDataRequest: video: 5gZOYKHXwyQ took: 414ms
D/Extended: StreamingDataRequest: Fetching video streams for: 5gZOYKHXwyQ using client: ANDROID_UNPLUGGED
D/Extended: StreamingDataRequest: video: 5gZOYKHXwyQ took: 206ms
D/Extended: SpoofStreamingDataPatch: Overriding video stream: 5gZOYKHXwyQ
```

I've not tested the private kids video, though.

However, it seems that Android TV cannot play copyrighted music(?). (I don't know the exact condition.)
So, if the video is "kids video" and "copyrighted music", all clients other than logged out iOS are not available.

https://www.youtube.com/watch?v=e_04ZrNroTo is an example.

```
D/Extended: StreamingDataRequest: Fetching video streams for: e_04ZrNroTo using client: IOS
I/Extended: StreamingDataRequest: IOS not available with response code: 400 message: Bad Request
D/Extended: StreamingDataRequest: video: e_04ZrNroTo took: 63ms
D/Extended: StreamingDataRequest: Fetching video streams for: e_04ZrNroTo using client: ANDROID_VR
D/Extended: StreamingDataRequest: video: e_04ZrNroTo took: 328ms
D/Extended: StreamingDataRequest: Fetching video streams for: e_04ZrNroTo using client: ANDROID_UNPLUGGED
D/Extended: StreamingDataRequest: video: e_04ZrNroTo took: 164ms
I/Extended: StreamingDataRequest: Could not fetch any client streams
D/Extended: SpoofStreamingDataPatch: Not overriding streaming data (video stream is null): e_04ZrNroTo
```

By the way, Android TV doesn't seem to work unless you're logged in.
Since iOS works if you are not logged in, I think the two clients can cover each other's weaknesses.

PR #4020, opened 2024-11-29
feat: Add `Change icon` patch

oSumAtrIX:
# About

This patch is meant to replace the current branding patch for YouTube but also enable patching any apps icon.

# Details

The idea is to ask the user for adaptive icons by default, and a fallback full icon. As it turns out there's like 10 ways to create an icon on Android. How the patch works is, it reads the manifest which references the icon. The reference can be a png file, an XML file, a drawable, mipmap and more. In the case of it being a mipmap and the reference pointing to an adaptive icon, the patch can be applied as it currently is on YouTube. In the other cases, the icon is replaced with fullIcon.

The patch isn't done yet, and there's probably a couple considerations that haven't been made yet, but that'll be clear during the implementation. For example fullIcon is currently just one icon, but IIRC there can be a drawable icon for each density for example, so scenarios like this need to be covered by the patch.

# Todo

- Finish implementation
- Test

### DIFF
--- a/patches/api/patches.api
+++ b/patches/api/patches.api
@@ -1,3 +1,7 @@
+public final class app/revanced/patches/all/layout/branding/IconPatchKt {
+	public static final fun getChangeIconPatch ()Lapp/revanced/patcher/patch/ResourcePatch;
+}
+
 public final class app/revanced/patches/all/misc/activity/exportall/ExportAllActivitiesPatchKt {
 	public static final fun getExportAllActivitiesPatch ()Lapp/revanced/patcher/patch/ResourcePatch;
 }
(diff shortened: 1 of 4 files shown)


oSumAtrIX:
So there are two questions. Right now, users have to provide icons for all densities. This is a bit annoying. Technically, a single icon can be provided, which Android the scales. However, due to the scaling, these tiny artifacts (such as aliasing) could look weird.
Additionally, the ability to provide adaptive icons is provided. Without those, Android can't use the FG or BG accordingly, such as for a splash screen. So, 3 icons for each density (monochrome exists, too). https://icon.kitchen/ is linked to the options, but it is still tedious.

@LisoUseInAIKyrios, any suggestions?

PR #4056, opened 2024-12-06
fix(YouTube Music - GmsCore support): Resolve patching errors

LisoUseInAIKyrios:
Fixes patching errors caused by extension refactoring


patches/src/main/kotlin/app/revanced/patches/music/misc/extension/SharedExtensionPatch.kt
```diff
@@ -4,6 +4,6 @@ import app.revanced.patches.music.misc.extension.hooks.applicationInitHook
 import app.revanced.patches.shared.misc.extension.sharedExtensionPatch
 
 val sharedExtensionPatch = sharedExtensionPatch(
-    "music",
+    "shared",
```
oSumAtrIX, line 7: There is an overload of this function without two parameters

oSumAtrIX (approved):

PR #4065, opened 2024-12-07
feat(YouTube Music): Add `Spoof video streams` patch to fix playback

oSumAtrIX:
image

This PR refactors the existing patch into a shared patch. YT Music does not have settings, so only the "default" settings can be utilized in the patch in YT Music until settings are added. 

While refactoring I noticed some inconsistencies in the extensions. Some classes are in a "patches" package, some are not. In the extensions the spoof streams patch is also present under a spoof package, in the patches it was under "fix/playback"

The patch is currently called "Spoof **video** streams" however, also audio streams are spoofed. A new name should be chosen.

The patch has not been tested extensively yet. I'll update the PR once some people confirm it works.


oSumAtrIX:
YT Music also has video playback, so "Spoof playback streams" works the best, however it's a bit more difficult to understand what it does now to the layman

oSumAtrIX:
> Looks good.
> 
> What to do about YT Music nagging its out of date? Delete the broken `Permanent shuffle` patch in a separate PR? It's compatible version is holding back the overall recommended version, and the patch no longer works with the newer releases and it's been broken a while.

The patch is disabled by default. ReVanced Manager is still calculating it with the rest of the patches to recommend a mutual version. There is an open issue to fix that.

Karthik-Subbu:
> Looks good.
> 
> What to do about YT Music nagging its out of date?  Delete the broken `Permanent shuffle` patch in a separate PR?  It's compatible version is holding back the overall recommended version, and the patch no longer works with the newer releases and it's been broken a while.

Can version info be spoofed to a higher version to remove the update nag when older versions are used for a particular feature?

Karthik-Subbu:
I was having a thought, can debug logging for yt music be added similar to yt? It may be helpful to address playback issues and may also assist in fixing broken patches (not sure though).

kairusds:
this patch seems to be working great but it also added audio normalisation which wasn't in the original app

kairusds:
> Is stable volume on by default?

most likely but there's no option like that in the yt music app itself

PR #4099, opened 2024-12-10
feat(Twitch): Make patches compatible with latest versions

oSumAtrIX:
This fixes patching errors, however late versions of Twitch are only bundles and the last ones that are APK fail to compile due to resources being decoded incorrectly. Neither of which is matter for the patches though.

### DIFF
--- a/patches/src/main/kotlin/app/revanced/patches/twitch/ad/embedded/Fingerprints.kt
+++ b/patches/src/main/kotlin/app/revanced/patches/twitch/ad/embedded/Fingerprints.kt
@@ -4,6 +4,6 @@ import app.revanced.patcher.fingerprint
 
 internal val createsUsherClientFingerprint = fingerprint {
     custom { method, _ ->
-        method.definingClass.endsWith("Ltv/twitch/android/network/OkHttpClientFactory;") && method.name == "buildOkHttpClient"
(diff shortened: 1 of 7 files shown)


patches/src/main/kotlin/app/revanced/patches/twitch/ad/embedded/Fingerprints.kt
```diff
@@ -4,6 +4,6 @@ import app.revanced.patcher.fingerprint
 
 internal val createsUsherClientFingerprint = fingerprint {
     custom { method, _ ->
-        method.definingClass.endsWith("Ltv/twitch/android/network/OkHttpClientFactory;") && method.name == "buildOkHttpClient"
+        method.name == "buildOkHttpClient" && method.definingClass.endsWith("Ltv/twitch/android/network/OkHttpClientFactory;")
```
kitadai31, line 7: Is it OK that it is not refactored to use classdef?
(I see you ReVanced team are always doing this)

Sbmhawk:
What version of twitch should I be using now?

sebm253:
@LisoUseInAIKyrios this compatibility array here wasn't changed so it's still limited to <16.9.1, is it meant to be like that? https://github.com/ReVanced/revanced-patches/blob/ea90c57d23816006f53702b284fd264a0462cbc1/patches/src/main/kotlin/app/revanced/patches/twitch/ad/video/VideoAdsPatch.kt#L158-#L162

oSumAtrIX:
That should be removed

PR #4113, opened 2024-12-13
fix(YouTube - Hide ads): Hide new type of featured promotions

### LINKED ISSUE #4111: feat(YouTube): Hide YouTube featured section
### Feature description

YouTube has this random recommendation that pop ups in home feed at times. Adding `brand_video_singleton.eml` should probably do it, as the section don't appear very often.

Image

### Motivation

It's either some content from YouTube's end or some promotion based content, also from YouTube. So it's better off hidden.

### Acknowledgements

ILoveOpenSourceApplications:
Closes #4111 .


ILoveOpenSourceApplications:
I think for now this will do, in the future if YouTube decides to use this for something else we can separate to its own option.

PR #4115, opened 2024-12-14
fix(YouTube Music - Spoof video streams): Use `IOS_MUSIC` client

oSumAtrIX:
Instead of the VR client, iOS Music is now used. This confirmed fixing playback issues for VR client users. No exceptions are logged using the VR client, so the playback itself failed for some reason. Using the iOS Music client type fixes the issue. It is unclear if other side effects exist such as stable volume.


extensions/shared/library/src/main/java/app/revanced/extension/shared/spoof/ClientType.java
```diff
@@ -45,61 +45,48 @@ public enum ClientType {
...
-    ANDROID_VR_NO_HL(
-            ANDROID_VR.id,
-            ANDROID_VR.clientName,
-            ANDROID_VR.deviceModel,
-            ANDROID_VR.osVersion,
-            ANDROID_VR.userAgent,
-            ANDROID_VR.androidSdkVersion,
-            ANDROID_VR.clientVersion,
-            ANDROID_VR.canLogin,
+    IOS_MUSIC(26,
```
oSumAtrIX, line 48: This may need logic to skip when using the YT client for example.

oSumAtrIX:
Seems like the issue still occurs, further debugging will be needed

extensions/shared/library/src/main/java/app/revanced/extension/shared/spoof/ClientType.java
inotia00: IOS_MUSIC client supports login
(canLogin = true)

So it can pass verification for age-restricted or private videos, and you don't need to set the audio track language either

(and as far as I know YouTube Music doesn't support audio tracks)

At least for YouTube Music we can play all videos with just the IOS_MUSIC client

oSumAtrIX:
Superseded by https://github.com/ReVanced/revanced-patches/pull/4132

Issue #1491, opened 2023-01-03
feat(YouTube Music): Black theme for cast pop-up panel


ILoveOpenSourceApplications:
### Application

YouTube Music

### Issue

As the title suggests, the cast option in YouTube Music is grey in colour while the entire other UI element is amoled black. A patch to make this theming uniform would be nice.

### Patch

An option or just making it black by default or a theme option in YouTube music.

### Motivation

The same way as everything is black, this would further improve upon that purpose if implemented.

### Acknowledgements

- [X] I have searched the existing issues and this is a new and no duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this issue properly.

ILoveOpenSourceApplications:
Yes.

Issue #1391, opened 2023-01-30
feat(YouTube Music): Theming & Customisation


ILoveOpenSourceApplications:
### Application

YT Music

### Issue

YT Music doesn't have many theming or customisation patches apart from the few basic patches to get the app ad free and running in the background. It would be nice for YT Music to get the same amount of attention YouTube app is getting.

### Patch

1. Black color for the navigation bar for a more uniform looking UI in the home screen.
2. Black color added to the casting panel which open up when pressing the cast icon.
3. Options to customise the name of the app to just be YT Music or YT Music ReVanced or whatever just like YouTube app.
4. Option to change icon (though there's no ReVanced music icon atm this can wait or just use a community made icon)
5. An option to change the minimized player color to follow the theme of the music playing in full screen.

### Motivation

To have YT Music to be treated in somewhat similar manner to that of YouTube app itself and be given the freedom and options to decide how the app looks and feels. Also, some of these features are present in RVX, which also motivated me to see the ReVanced version to have these be implemented in its version also.

### Acknowledgements

- [X] I have searched the existing issues and this is a new and no duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this issue properly.

Issue #614, opened 2023-07-12
feat(YouTube Music): Theming & customization options


ILoveOpenSourceApplications:
### Application

YT Music 

### Issue

Currently, YT Music lacks several customization features that could greatly enhance user experience and personalization within the music streaming platform. I would like to propose the implementation of several theming and customization options for YT Music, similar to the features available in the YouTube app.

### Patch

- [ ] `Amoled Black Theme`:

Currently, YT Music features elements in gray color. It would be great to have an option to change these elements to black, creating an overall black theme. This will provide a sleek and visually appealing interface for users who prefer darker themes.

- [ ] `Customizable App Name`:

Provide the ability to customize the name of the app, similar to the YouTube app. Users should be able to change the app name to "YT Music" or "YT Music ReVanced" or any other desired name. This customization feature adds a personal touch and aligns with the user's preferences.

- [ ] `Icon Customization`:

Although there is currently no ReVanced music icon available, it would be beneficial to include an option to change the app icon in the future. Users can then choose from a selection of community-made icons or even upload their own icon to further personalize their YT Music experience.

- [ ] `Minimized Player Color`:

Introduce an option to customize the color of the minimized player to match the theme of the music playing in full-screen mode. This feature ensures consistency in the user interface and provides a seamless transition between different views.

- [ ] `Remember Video Quality`:

Enable YT Music to remember the preferred video quality for songs. This will allow users to set their desired video quality once and have it applied automatically to future song playback. It saves users time and ensures they always enjoy their preferred video quality.

- [ ] `Hide New Playlist Button in Library`:

Add an option to hide the "New Playlist" button in the library section. This feature would be beneficial for users who prefer to manage their playlists in other ways and want a cleaner interface without unnecessary buttons.

- [ ] `Disable Auto Captions`:

Provide an option to disable auto-generated captions in video songs. Some users may find the auto captions distracting or prefer to view the content without captions, and having the ability to disable them would enhance their viewing experience.

- [ ] `Enable Tablet/Landscape Mode`:

Introduce tablet and landscape mode support for YT Music. This feature is particularly important for users who frequently use their tablets or prefer landscape orientation on larger screens. It will provide a more immersive and enjoyable experience for such users.

### Motivation

By implementing these theming and customization options, YT Music will offer users greater control over their experience and allow them to tailor the app according to their preferences. These enhancements will contribute to a more personalized and enjoyable music streaming experience overall.

### Acknowledgements

[...]

PR #4258, opened 2025-01-03
feat(YouTube Music - AMOLED black theme): Add `AMOLED black theme` patch

### LINKED ISSUE #609: feat (YT Music/`Amoled Black Theme`): Forces Amoled Black Theme throughout the YT Music UI
### Application

YT Music 

### Issue

Currently, YT Music lacks several customization features that could greatly enhance user experience and personalization within the music streaming platform. I would like to propose the implementation of several theming and customization options for YT Music, similar to the features available in the YouTube app.

### Patch

Currently, YT Music features elements in gray color. It would be great to have an option to change these elements to black, creating an overall black theme. This will provide a sleek and visually appealing interface for users who prefer darker themes.

### Motivation

By implementing these theming and customization options, YT Music will offer users greater control over their experience and allow them to tailor the app according to their preferences. These enhancements will contribute to a more personalized and enjoyable music streaming experience overall.

### Acknowledgements

ILoveOpenSourceApplications:
Closes #609, #614 (partially), #1391(partially), and #1491. (I know, all of them are my issues, crazy right.)

I have no idea what I've done here. But, since there's no one else to do this (at least for now), I shall be the Guinea pig. Open to changes (and many of that shall come).

Credits: @inotia00


extensions/music/src/main/java/app/revanced/extension/music/ThemeHelper.java
oSumAtrIX, line 1: The file should be moved to the shared extension

patches/src/main/kotlin/app/revanced/patches/music/layout/amoledblacktheme/AmoledBlackThemePatch.kt
```diff
@@ -0,0 +1,71 @@
+package app.revanced.patches.music.layout.amoledblacktheme
+
+import app.revanced.patcher.extensions.InstructionExtensions.addInstructions
+import app.revanced.patcher.patch.bytecodePatch
+import app.revanced.patcher.patch.resourcePatch
+import app.revanced.patcher.util.smali.toInstructions
+import org.w3c.dom.Element
+
+val amoledBlackThemePatch = bytecodePatch(
```
oSumAtrIX, line 9: This patch already exists for YouTube and should be abstracted if both apps need specific implementations

oSumAtrIX (changes requested):

extensions/music/src/main/java/app/revanced/extension/music/ThemeHelper.java
ILoveOpenSourceApplications: The current one is in YT's directory which is why I created a separate one for YTM. So how do you propose the change to be?

oSumAtrIX: It needs to be moved into the shared package, similar to the other shared patches

ILoveOpenSourceApplications: To transfer `ThemeHelper.java` to the shared path folder?

oSumAtrIX: The point is to reduce duplicate code via abstraction

patches/src/main/kotlin/app/revanced/patches/music/layout/amoledblacktheme/AmoledBlackThemePatch.kt
ILoveOpenSourceApplications: I rebased the patch based on the current YouTube theme patch. So, should I create a single theme patch across both? I don't know what I'm about to do is the right step.

oSumAtrIX: > So, should I create a single theme patch across both

Thats the idea

extensions/music/src/main/java/app/revanced/extension/music/ThemeHelper.java
ILoveOpenSourceApplications: But YTM isn't as flexible in the theming department as there's only a couple of elements. So, should we go with approach?

ILoveOpenSourceApplications:
Is it any better now or just worse?

oSumAtrIX:
You're getting there; however, there's missing/incorrect abstraction, which isn't as easy to pull off as copying or moving files. I think it's better leaving this to us.

ILoveOpenSourceApplications:
@oSumAtrIX , do you think navigation bar theme is defined in the xml files so that this can work ?
```kotlin
document("res/values/themes.xml").use { document -> val resourcesNode = document.getElementsByTagName("resources").item(0) as Element for (i in 0 until resourcesNode.childNodes.length) { val node = resourcesNode.childNodes.item(i) as? Element ?: continue if (node.getAttribute("name") == "navigation_bar_color") { node.textContent = amoledBlackColor
```

ILoveOpenSourceApplications:
The reason I asked is mentioned in this https://github.com/ReVanced/revanced-patches/issues/609#issuecomment-2557253704. As currently, ReVanced YouTube Music settings page implementation is in works, only features that can be achieved while patching can be added, which is the theme patch. 

Theming should all fall under one patch and the nav bar  theme being separate felt off to me and I would like it to come under this patch as well but I don't know how to achieve that. Will you be able to look into that as well or should I do something about that?

oSumAtrIX:
I don't know what you're referring to with the settings, but theming for YT doesn't rely on the settings patch.

ILoveOpenSourceApplications:
> I don't know what you're referring to with the settings, but theming for YT doesn't rely on the settings patch.

I don't think you read my comment completely.

ILoveOpenSourceApplications:
Please refer [patches/src/main/kotlin/app/revanced/patches/music/layout/theme/DarkThemePatch.kt](https://github.com/inotia00/revanced-patches/blob/1aaf03df50b6d3d4b85e3ecac07442a7eb77a839/patches/src/main/kotlin/app/revanced/patches/music/layout/theme/DarkThemePatch.kt) as he has made some changes and is now much more similar to our YouTube `Theme` patch.

PR #4274, opened 2025-01-06
chore: Clarify reddit Spoof client app instructions

VasilisThePikachu:
With the kebab menu on the side asking for a file/folder. It can be easily misunderstood by users that this is a file selection prompt. When it actually is a string field asking for the client ID.

This can be even more confusing to users who have their client id inside the ``reddit_client_id_revanced.txt`` file from previous version of this patch. (and from various out of date online instructions)

Preferably, this would be fixed in the app by removing the Kebab menu or making it input the file contents instead. But I am not an android dev.

Note: IDK if this is a "chore" I never experienced pr naming guidelines like these.


oSumAtrIX:
Patches have no context to ReVanced Manager or it's implementation. ReVanced CLI is a command line application where you input options in the command line and not text boxes, so it would evidently be wrong for the patch description to mention input boxes. If it were to ask for a file, the description would explicitly mention that it asks for a file. A client id is not a file. A file containing the client id is. The option asks for a client id and not a file containing the client id. Thanks regardless

PR #4368, opened 2025-01-29
feat(nu.nl): add remove ads patch

jabbink:
Adds a patch to remove several entry points for ads in `nl.sanomamedia.android.nu`


patches/src/main/kotlin/app/revanced/patches/nunl/ads/Fingerprints.kt
```diff
@@ -0,0 +1,52 @@
...
+
+import app.revanced.patcher.fingerprint
+import com.android.tools.smali.dexlib2.AccessFlags
+import com.android.tools.smali.dexlib2.Opcode
+
+internal val xandrInitFingerprint = fingerprint {
+    accessFlags(AccessFlags.PUBLIC, AccessFlags.STATIC)
+    returns("V")
+    custom { methodDef, classDef ->
+        classDef.type == "Lcom/appnexus/opensdk/XandrAd;" && methodDef.name == "init" && methodDef.parameters.size == 4
```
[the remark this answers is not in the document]
jabbink: I was about to change the fingerprint, but tried finding these "sponsored articles" (it's a news app showing these adverts inline as if it's news) without the patch enabled, and they seem to be gone now so I cannot verify if this patch was actually correct in the first place and I fully removed it for now

patches/src/main/kotlin/app/revanced/patches/nunl/ads/HideAdsPatch.kt
```diff
@@ -0,0 +1,47 @@
...
+                move-result-object v0
+                return-object v0
+            """,
+        )
+
+        // filter injected content from API calls out of lists
+        arrayOf(screenMapperFingerprint, nextPageRepositoryImplFingerprint).forEach {
+            val startIndex = it.patternMatch!!.startIndex
+            it.method.apply {
+                val moveInstruction = getInstruction<Instruction11x>(startIndex)
```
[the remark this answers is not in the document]
jabbink: adjusted to `OneRegisterInstruction`

extensions/nunl/src/main/java/app/revanced/extension/nunl/ads/HideAdsPatch.java
```diff
@@ -0,0 +1,103 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+import app.revanced.extension.shared.Logger;
+
+@SuppressWarnings("unused")
+public class HideAdsPatch {
+    private static final String[] blockedHeaderBlocks = {
+            "Aanbiedingen (Adverteerders)",
```
oSumAtrIX, line 14: Is the string always the same regardless of system language?

extensions/nunl/src/main/java/app/revanced/extension/nunl/ads/HideAdsPatch.java
```diff
@@ -0,0 +1,103 @@
...
+                            }
+                        }
+                        if (skipFullHeader) {
+                            index++;
+                            continue;
+                        }
+                    }
+                }
+
+                if (!skipFullHeader && !skipUntilDivider) {
```
oSumAtrIX, line 90: My IDE shows that this check can be removed. Is this actually the case?

patches/src/main/kotlin/app/revanced/patches/nunl/ads/HideAdsPatch.kt
```diff
@@ -0,0 +1,47 @@
...
+                invoke-direct { v0 }, Lcom/jwplayer/pub/api/configuration/ads/VastAdvertisingConfig${'$'}Builder;-><init>()V
+                invoke-virtual { v0 }, Lcom/jwplayer/pub/api/configuration/ads/VastAdvertisingConfig${'$'}Builder;->build()Lcom/jwplayer/pub/api/configuration/ads/VastAdvertisingConfig;
+                move-result-object v0
+                return-object v0
+            """,
+        )
+
+        // Filter injected content from API calls out of lists.
+        arrayOf(screenMapperFingerprint, nextPageRepositoryImplFingerprint).forEach {
+            val startIndex = it.patternMatch!!.startIndex
```
oSumAtrIX, line 32: Maybe label the variable so it is understandable what that index is. Is it an index where blocks are added to the UI?

patches/src/main/kotlin/app/revanced/patches/nunl/ads/HideAdsPatch.kt
```diff
@@ -0,0 +1,47 @@
...
+val hideAdsPatch = bytecodePatch(
+    name = "Hide ads",
+    description = "Hide ads and sponsored articles in list pages and remove pre-roll ads on videos.",
+) {
+    compatibleWith("nl.sanomamedia.android.nu"("11.0.0", "11.0.1"))
+
+    extendWith("extensions/nunl.rve")
+
+    execute {
+        // Disable video pre-roll ads.
```
oSumAtrIX, line 18: Comment the strategy of this patch. It seems like you replace a configuration with an empty one which disables the ads?

extensions/nunl/src/main/java/app/revanced/extension/nunl/ads/HideAdsPatch.java
jabbink: It is independent from system language, but sadly it comes from an online GQL endpoint and can change at any time. From what I saw, no other flags indicate that everything following these texts are spam, until a next "Header" is mentioned.

extensions/nunl/src/main/java/app/revanced/extension/nunl/ads/HideAdsPatch.java
jabbink: I have the same (JetBrains?), but it's false. It assumed the stubbed methods never return anything and throw an exception instead, so the conditionals further above will never set either `skipFullHeader` or `skipUntilDivider`.

patches/src/main/kotlin/app/revanced/patches/nunl/ads/HideAdsPatch.kt
jabbink: adjusted

patches/src/main/kotlin/app/revanced/patches/nunl/ads/HideAdsPatch.kt
jabbink: added explanation

patches/src/main/kotlin/app/revanced/patches/nunl/firebase/Fingerprints.kt
```diff
@@ -0,0 +1,14 @@
...
+import app.revanced.patcher.fingerprint
+import com.android.tools.smali.dexlib2.AccessFlags
+
+internal val getFingerprintHashForPackageFingerprint = fingerprint {
+    accessFlags(AccessFlags.PRIVATE)
+    parameters()
+    returns("Ljava/lang/String;")
+
+    custom { methodDef, classDef ->
+        classDef.type.startsWith("Lcom/google/firebase/") && methodDef.name == "getFingerprintHashForPackage"
```
oSumAtrIX, line 12: classDef already extends CharSequence, you can just call startsWith on it

jabbink: This was always intended to match 3 methods; adjusted to now have 3 signatures in separate classes and patch all 3

still relies on `classDef.type` now to check for equality of full class name, as many other fingerprints in this project do  
that said, the [documentation](https://github.com/ReVanced/revanced-patcher/blob/main/docs/2_2_1_fingerprinting.md#%EF%B8%8F-example-fingerprint) implies that one could check that as well without the `.type` which I think is incorrect or I'm doing something wrong

extensions/nunl/src/main/java/app/revanced/extension/nunl/ads/HideAdsPatch.java
```diff
@@ -0,0 +1,114 @@
...
+                    cleanedList.add(currentBlock);
+                }
+                index++;
+            }
+
+            // Replace list in-place to not deal with moving the result to the correct register in smali.
+            blocks.clear();
+            blocks.addAll(cleanedList);
+        } catch (Exception ex) {
+            Logger.printException(() -> "filterAds failure", ex);
```
jabbink, line 111: This crashes by the way

```
02-21 07:00:45.244 14973 14973 E AndroidRuntime: java.lang.ExceptionInInitializerError
02-21 07:00:45.244 14973 14973 E AndroidRuntime:        at app.revanced.extension.shared.settings.BaseSettings.<clinit>(BaseSettings.java:18)
02-21 07:00:45.244 14973 14973 E AndroidRuntime:        at app.revanced.extension.shared.Logger.printException(Logger.java:134)
[HideAdsPatch stacktrace into original app]
02-21 07:00:45.244 14973 14973 E AndroidRuntime: Caused by: java.lang.NullPointerException
02-21 07:00:45.244 14973 14973 E AndroidRuntime:        at java.util.Objects.requireNonNull(Objects.java:207)
02-21 07:00:45.244 14973 14973 E AndroidRuntime:        at app.revanced.extension.shared.settings.preference.SharedPrefCategory.<init>(SharedPrefCategory.java:30)
02-21 07:00:45.244 14973 14973 E AndroidRuntime:        at app.revanced.extension.shared.settings.Setting.<clinit>(Setting.java:101)
02-21 07:00:45.244 14973 14973 E AndroidRuntime:        ... 15 more
```

Briefly looking through other extensions using `Logger.printException` I see the extension of `BaseSettings` and a menu to actually toggle these settings, which I was not planning to implement for this. I am not familiar enough with this project to quickly figure out how to just set this `DEBUG_TOAST_ON_ERROR` to a fixed value to prevent a crash in the error handler.

jabbink: > hook the main activity onCreate method 

thanks, that worked and the toast is now shown in case an exception is thrown

patches/src/main/kotlin/app/revanced/patches/nunl/ads/HideAdsPatch.kt
```diff
@@ -0,0 +1,51 @@
...
+                invoke-direct { v0 }, Lcom/jwplayer/pub/api/configuration/ads/VastAdvertisingConfig${'$'}Builder;-><init>()V
+                invoke-virtual { v0 }, Lcom/jwplayer/pub/api/configuration/ads/VastAdvertisingConfig${'$'}Builder;->build()Lcom/jwplayer/pub/api/configuration/ads/VastAdvertisingConfig;
+                move-result-object v0
+                return-object v0
+            """,
+        )
+
+        // Filter injected content from API calls out of lists.
+        arrayOf(screenMapperFingerprint, nextPageRepositoryImplFingerprint).forEach {
+            // index of instruction moving result of BlockPage;->getBlocks(...)
```
oSumAtrIX, line 34: Comments start with a big letter and end with a dot.

oSumAtrIX (approved):
LGTM once the last review is resolved.

PR #4438, opened 2025-02-11
fix(YouTube Music): update `Enable exclusive audio playback` patch to be more reliable across multiple versions

ponces:


ponces:
@LisoUseInAIKyrios can you merge this too please?

Issue #521, opened 2023-08-05
bug(Twitter / X - Hide recommended users): `Who to Follow` is not hiding


kazimmt:
### Type

Other

### Bug description

Who to Follow in twitter is not hiding in last few twitter version.

### Steps to reproduce

- Patch Twitter
- Install it
- Login
- Scroll down in `For You` Tab & `Following` tab

### Screenshots or videos

![-2147483648_-210286](https://github.com/ReVanced/revanced-integrations/assets/82371061/89a43d81-e3b4-4184-94aa-99fc60088e73)

kazimmt:
> Is this issue still present?

Yes, it is.

kazimmt:
![Screenshot_20240506-203916_X](https://github.com/ReVanced/revanced-patches/assets/82371061/527c3227-5433-4bb9-8203-b6c7f3f36d37)

PR #4456, opened 2025-02-13
fix(Twitter - Hide recommended users): add new timeline entryId prefix

trogper:
twitter has started (a long ago) to use `who-to-subscribe-` prefix instead of `who-to-follow-`

this adds the new prefix, so it is properly blocked again

[!fixes ](https://github.com/ReVanced/revanced-patches/issues/521)

### DIFF
--- a/extensions/twitter/src/main/java/app/revanced/twitter/patches/hook/twifucker/TwiFucker.kt
+++ b/extensions/twitter/src/main/java/app/revanced/twitter/patches/hook/twifucker/TwiFucker.kt
@@ -163,7 +163,7 @@ internal object TwiFucker {
 
     private fun JSONObject.entryIsWhoToFollow(): Boolean =
         optString("entryId").let {
-            it.startsWith("whoToFollow-") || it.startsWith("who-to-follow-") || it.startsWith("connect-module-")
+            it.startsWith("whoToFollow-") || it.startsWith("who-to-follow-") || it.startsWith("connect-module-") || it.startsWith("who-to-subscribe-")
         }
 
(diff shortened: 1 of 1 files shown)


trogper:
this actually won't work on newer versions of twitter apk, since TwiFucker used to hook onto org.json and they replaced org.json library with LoganSquare.

trogper:
I am not sure. I tried adding logging to the hook, but as I later found out, that won't work because of the library replacement.

trogper:
According to issue 521, it stopped working before August of 2023. But that may also be caused by change of the prefix.

The folks at [crimera/piko](https://github.com/crimera/piko) have working twitter patches, but they are based on older versions of revanced patch API, so it's incompatible with current revanced cli/app for now.

extensions/twitter/src/main/java/app/revanced/twitter/patches/hook/twifucker/TwiFucker.kt
```diff
@@ -163,7 +163,7 @@ internal object TwiFucker {
 
     private fun JSONObject.entryIsWhoToFollow(): Boolean =
         optString("entryId").let {
-            it.startsWith("whoToFollow-") || it.startsWith("who-to-follow-") || it.startsWith("connect-module-")
+            it.startsWith("whoToFollow-") || it.startsWith("who-to-follow-") || it.startsWith("connect-module-") || it.startsWith("who-to-subscribe-")
```
oSumAtrIX, line 166: Is who-to-follow still required?

PR #4537, opened 2025-03-04
Add max volume customization patch

### LINKED ISSUE #1646: bug: In certain devices, gesture volume goes up to 150, instead of 15.
### Type

Other

### Bug description

On certain devices (I have Redmi Note 10 Pro), the volume with gestures enabled goes up to level 150 instead of 15. The drag speed is also same as level 15, so in one drag maximum volume I can change is 30. That's 5 full drags to change from min to max volume and max to min.

### Steps to reproduce

1. Enable gesture for brightness and volume

2. Use volume gesture to increase or decrease volume

3. Instead going through level 0 through 15, it goes 0 through 15 with same slow drag speed.

### Relevant log output

```shell
sorry not available
```

### Screenshots or videos

_No response_

### Solution

Currently it's 100 levels for brightness with high drag sensitivity, and 15 levels for volume with low drag sensitivity.

I think it's fair to do 30 levels of brightness and volume in all devices, as in MX Player, (or 100 as in VLC Android) for the same drag sensitivity.

### Additional context

_No response_

### Acknowledgements

xVemu:
It's super annoying to scroll through 160 volume levels, so I've added an option to customize the number of volume levels when swiping.

Fixes #1646

Btw, `audioManager.getStreamMaxVolume(targetStream)` returns different values on my device. With root/mount method it returns 160, but with the microg method, it's just 16.

### DIFF
--- a/extensions/youtube/src/main/java/app/revanced/extension/youtube/settings/Settings.java
+++ b/extensions/youtube/src/main/java/app/revanced/extension/youtube/settings/Settings.java
@@ -315,6 +315,7 @@ public class Settings extends BaseSettings {
             parentsAny(SWIPE_BRIGHTNESS, SWIPE_VOLUME));
     public static final IntegerSetting SWIPE_MAGNITUDE_THRESHOLD = new IntegerSetting("revanced_swipe_threshold", 30, true,
             parentsAny(SWIPE_BRIGHTNESS, SWIPE_VOLUME));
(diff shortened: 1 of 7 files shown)


oSumAtrIX:
Not sure why mounting should make a difference, you can open an issue regarding that

xVemu:
I will do another PR with sensitivity approach, because I think it's better

pmsobrado:
I would also like an option to modify volume steps. My Samsung phone has 150 volume levels, and the side buttons increase them by 5, this is, 30 steps. But swiping on Revanced increases the volume by 10, this is, 15 steps. I would like to have 30 or even 60 volume steps in Revanced swipe function, thanks.

PR #4543, opened 2025-03-05
feat(YouTube - Remember video quality): Separate settings for Shorts

aliernfrog:
Adds options to use separate quality preferences for Shorts.

This also lets users disable the patch specifically for Shorts.

## All changes
- Added a new "Video quality" preference screen to video settings
- Added 3 options to the new video quality preference screen:
  - Separate Shorts quality settings: when disabled (default), both Shorts and regular videos use the regular preferences
  - Remember Shorts quality changes
  - Default Shorts quality on Wi-Fi network
  - Default Shorts quality on mobile network
- Moved default video quality options and "remember video quality changes" switch to the new video quality preference screen

Screenshot_20250305-143110_YouTube
Screenshot_20250305-143120_YouTube

### DIFF
--- a/extensions/youtube/src/main/java/app/revanced/extension/youtube/patches/playback/quality/RememberVideoQualityPatch.java
+++ b/extensions/youtube/src/main/java/app/revanced/extension/youtube/patches/playback/quality/RememberVideoQualityPatch.java
@@ -12,15 +12,20 @@ import java.util.List;
 
 import app.revanced.extension.shared.Logger;
 import app.revanced.extension.shared.Utils;
+import app.revanced.extension.shared.settings.BooleanSetting;
 import app.revanced.extension.shared.settings.IntegerSetting;
 import app.revanced.extension.youtube.patches.VideoInformation;
 import app.revanced.extension.youtube.settings.Settings;
+import app.revanced.extension.youtube.shared.PlayerType;
 
 @SuppressWarnings("unused")
 public class RememberVideoQualityPatch {
     private static final int AUTOMATIC_VIDEO_QUALITY_VALUE = -2;
     private static final IntegerSetting wifiQualitySetting = Settings.VIDEO_QUALITY_DEFAULT_WIFI;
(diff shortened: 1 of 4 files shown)


aliernfrog:
> `Separate Shorts quality settings` might be removed for simplicity.

That would work too, but the Shorts check is just a `PlayerType.isNoneHiddenOrSlidingMinimized()` so it can sometimes wrongfully apply Shorts quality to a regular video, which can be annoying for some users and this toggle minimizes the chances of this. 

> Why not keep all of these settings in the Video menu, with no submenu?  A submenu makes another step to navigate to the video quality settings.
In the Video menu, sorting depends on the language and can make the options of this patch all over the place. However, if the `Separate Shorts quality settings` toggle is removed, the sorting wouldn't matter that much.

aliernfrog:
> @aliernfrog Can you reproduce this?
> 
> - Change Shorts quality to 144p
> - Open a Short, notice the quality is set to 144p (or 240p if 144p is not available) - Correct
> - Close the Short
> - Change Shorts default quality to Auto
> - Open a different Short
> - Notice quality is still stuck at 144p or 240p - Incorrect
> 
> This does not happen with regular videos.

I think that is the intended behavior of YouTube, even without setting a default quality and setting current Shorts video to 144p, the quality is remembered until you restart the app.

Restarting the app automatically sounds good.

aliernfrog:
> `Restore old video quality menu` could be renamed.  It's been ages (5 years?) since the 'old' menu was present, and I don't think many users understand what the old menu was, or why the 'old' is desirable.
> 
> What the patch now does, is always opens the advanced quality menu.  So naming the patch something like `Advanced video quality menu` or something similar would be much easier to understand what the patch does and why it's desirable.
> 
> 
> Or maybe better, merge `Remember video quality` together with `Restore old video quality` and maybe also `Disable HDR` into a single patch: `Video quality`.  Then it's similar to `Playback speed` that has multiple speed related functions.

Merging those patches is a good idea. `Remember video quality` patch alone does not provide a way to select a specific quality while playing Shorts, and some users may get confused when they have this patch enabled but not the `Old video quality menu`.

PR #4551, opened 2025-03-06
fix(Boost for Reddit - Spoof client ID): Add additional patch to spoof Boost for Reddit's user agent and redirect URI

### LINKED ISSUE #4549: bug: Boost for Reddit 403 errors
### Bug description

Boost for Reddit now gives a 403 error.  While I'm not certain it is most likely the User-Agent as there is a post to redditdev about this:

  * https://old.reddit.com/r/redditdev/comments/1j2sgxw/please_ensure_your_useragents_are_unique_and/

I can confirm using apktool modifying the `User-Agent` fixes the problem.

There is a suggested fix below: https://github.com/ReVanced/revanced-patches/issues/4549#issuecomment-2702891927 from @wchill 

### Error logs

```shell
There are none but Boost gives a response there is a `403` error from Reddit.
```

### Solution

The patch needs to be updated with new user input fields to allow customising the User-Agent as per Reddits requirements:

The best solution may just be to allow for a fully editable User-Agent with a suggestion of the format required by Reddit.

> Change your client's User-Agent string to something unique and descriptive, including the target platform, a unique application identifier, a version string, and your username as contact information, in the following format:
> `<platform>:<app ID>:<version string> (by /u/<reddit username>)`

### Additional context

I think what has happened is the API key for the old app has been pulled that is causing `403` in one case.  The ReVanced patch uses a random number for the platform which is now banned -- any set of random characters are banned and they have explicitly called out random integers which the ReVanced app uses. 

### Acknowledgements

wchill:
Fixes #4549.

This patch adds two new options `user-agent` and `redirect-uri` that can be used to spoof the user agent and OAuth redirect URI for Boost for Reddit.

It takes a different approach from the earlier user agent patch that used the old ReVanced API by directly replacing the format string used to construct the user agent.

Other clients will likely eventually need a similar patch; however, I don't have the time to implement this for other clients. A new interface has been added that other clients can move to as needed.

### DIFF
--- a/patches/api/patches.api
+++ b/patches/api/patches.api
@@ -421,6 +421,7 @@ public final class app/revanced/patches/reddit/customclients/FixSLinksPatchKt {
 
 public final class app/revanced/patches/reddit/customclients/SpoofClientPatchKt {
 	public static final fun spoofClientPatch (Ljava/lang/String;Lkotlin/jvm/functions/Function2;)Lapp/revanced/patcher/patch/BytecodePatch;
+	public static final fun spoofClientPatch (Lkotlin/jvm/functions/Function4;)Lapp/revanced/patcher/patch/BytecodePatch;
 	public static synthetic fun spoofClientPatch$default (Ljava/lang/String;Lkotlin/jvm/functions/Function2;ILjava/lang/Object;)Lapp/revanced/patcher/patch/BytecodePatch;
 }
 
--- a/patches/src/main/kotlin/app/revanced/patches/reddit/customclients/SpoofClientPatch.kt
+++ b/patches/src/main/kotlin/app/revanced/patches/reddit/customclients/SpoofClientPatch.kt
@@ -32,3 +32,51 @@ fun spoofClientPatch(
         ),
     )
 }
+
+/**
+ * Base class for patches that spoof the Reddit client.
+ *
+ * @param block The patch block. It is called with the client ID option, redirect URI
+ * option and user agent option.
+ */
+fun spoofClientPatch(
+    block: BytecodePatchBuilder.(
+        clientIdOption: Option<String>,
+        redirectUriOption: Option<String>,
+        userAgentOption: Option<String>,
+    ) -> Unit,
+) = bytecodePatch(
+    name = "Spoof client",
+    description = "Restores functionality of the app by using custom client ID.",
+) {
+    block(
+        stringOption(
+            "client-id",
+            null,
+            null,
+            "OAuth client ID",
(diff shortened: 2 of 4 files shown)


wchill:
I added a new interface that the other clients can use. As mentioned, I can't implement the actual patches for all the clients due to lack of time, so I'm hoping users of other clients can contribute as needed.

oSumAtrIX:
The user agent can be automatically generated by the patch if necessary. The redirect URI never needed to be changed, and this also should be the case for the user agent. Reddit is breaking their own API. Thanks regardless for the patch!

wchill:
Why has the PR been closed? The user agent being automatically generated by the patch is why users are running into issues. The redirect URI patch was added to reduce the chances of detection so we don't need to do this song and dance again in the future.

As explained [here](https://www.reddit.com/r/revancedapp/comments/1j4hene/reddit_boost_revanced_throwing_403_blocked_message/mg9uf6s/), reddit has some block on their side that is uniquely triggered by adding `rubenmayayo` anywhere in the user agent. Given how uniquely targeted the block is, it is better that it and the redirect URI be configurable so that reddit doesn't have a way to block Boost in the future.

oSumAtrIX:
Multiple reasons, the redirect uri does not need to change. Reddit does not "detect" third party clients. In fact they were never banned. Reddit merely monetized the API, which if you create your own client is free. Also a lot of copy paste requiring rewriting the PR anyways when just 3 lines need to change

wchill:
I'm not a Kotlin expert, so if someone else wants to improve the PR, they can. It's not a reason to close it. But the patches are clearly working for Boost users.

Reddit is clearly starting to crack down on 3rd party apps. I was not even using ReVanced to patch the Boost app previously because I had moderator status, so the app continued to work for me even after the API was monetized. That is, until today. There is no other reason for this app to just stop working when others are continuing to work.

oSumAtrIX:
I'll push a couple of lines changes myself rather than refactoring the entire PR. Reddit "cracking" down is just like i said them monetizing the API, nothing less nothing more.

oSumAtrIX:
For reference https://github.com/ReVanced/revanced-patches/commit/5d3c8175b34a3f6ae2732b25db0851773a8c000d

wchill:
Well, if you want to do it your way, fine, but this is going to be an easy target for later crackdowns from reddit (which is explicitly why I didn't do it your way).

oSumAtrIX:
For the third time, the "crackdown" is just the API being monetized. Reddit isn't going after oAuth applications.

Issue #3094, opened 2024-04-27
bug: The volume should be 1-2-3 instead of 10-20-30.


allahin:
### Bug description

The volume is increased out of 100, not out of 10, and it takes too long to increase the volume. I noticed that it doesn't happen on my Samsung device, but it happens on my Xiaomi device. My device: Xiaomi Redmi Note 12 Pro.

MIUI optimization is turned off.

### Solution

I think it has something to do with Xiaomi devices.

### Additional context

![Screenshot_2024-04-27-10-50-46-464_app revanced android youtube](https://github.com/ReVanced/revanced-patches/assets/166995789/936f5305-87c9-4574-8c61-124e601a4c51)

### Acknowledgements

- [X] This issue is not a duplicate of an existing bug report.
- [X] I have chosen an appropriate title.
- [X] All requested information has been provided properly.

oSumAtrIX:
That's the steps your phone reports. We could add a setting for swipe sensitivity though.

allahin:
You'd better do it because it takes a really long time to increase the volume.

Issue #1108, opened 2023-04-23
Volume is increasing by 1 instead of system level


saikrishna7149:
### Type

Other

### Bug description

Volume level is increasing step by step instead of system default step.

Example
Usually my volume increases by 10% in the stock yt
In patch it increases by 1%

### Steps to reproduce

Open yt patched and open video and increase volume.

Occurred in mi 11x 

### Relevant log output

```shell
na
```

### Solution

Use system level volume instead of yt volume level

### Acknowledgements

- [X] I have searched the existing issues and this is a new and no duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this issue properly.

PR #4557, opened 2025-03-06
feat(YouTube - Swipe controls): Add option to change volume swipe sensitivity

### LINKED ISSUE #1646: bug: In certain devices, gesture volume goes up to 150, instead of 15.

xVemu:
It's super annoying to scroll through 160 volume levels, so I've added an option to customize how much volume a single swipe changes.

May fix #1646, #3094, #1108

### DIFF
--- a/extensions/youtube/src/main/java/app/revanced/extension/youtube/swipecontrols/controller/gesture/core/BaseGestureController.kt
+++ b/extensions/youtube/src/main/java/app/revanced/extension/youtube/swipecontrols/controller/gesture/core/BaseGestureController.kt
@@ -24,6 +24,7 @@ abstract class BaseGestureController(
         controller.overlay,
         10,
         1,
+        controller.config.volumeSwipeMultiplier,
     ) {
 
     /**
--- a/extensions/youtube/src/main/java/app/revanced/extension/youtube/swipecontrols/controller/gesture/core/BaseGestureController.kt
+++ b/extensions/youtube/src/main/java/app/revanced/extension/youtube/swipecontrols/controller/gesture/core/BaseGestureController.kt
@@ -24,7 +24,7 @@ abstract class BaseGestureController(
         controller.overlay,
         10,
         1,
-        controller.config.volumeSwipeMultiplier,
+        controller.config.volumeSwipeSensitivity,
     ) {
 
     /**
--- a/patches/src/main/resources/addresources/values/strings.xml
+++ b/patches/src/main/resources/addresources/values/strings.xml
@@ -515,7 +515,7 @@ Adjust volume by swiping vertically on the right side of the screen"</string>
             <string name="revanced_swipe_threshold_title">Swipe magnitude threshold</string>
             <string name="revanced_swipe_threshold_summary">The amount of threshold for swipe to occur</string>
             <string name="revanced_swipe_volume_sensitivity_title">Volume swipe sensitivity</string>
-            <string name="revanced_swipe_volume_sensitivity_summary">Adjust volume change per swipe</string>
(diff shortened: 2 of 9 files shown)


xVemu:
> Is there ever a time when the swipe is too sensitive and the user wants swiping to change the volume less?

Currently, this would result in volume like 0.25, 0.67 etc. which could cause issues with changing volume.
I think there's no need for value smaller than 1.
Although 0 is ok since it just disables changing volume.

> Should swipe brightness also have a sensitivity setting?

I don't think it's necessary

xVemu:
> A multiplier/sensitivity of zero makes no sense and would make swiping do nothing. If volume swiping is to be turned off then turn it off with the enable volume swipe switch.

Yeah, I agree, but what should 0 do then? Act like 1?
I guess there's no way to set a minimum value of 1 in dialog?

patches/src/main/resources/addresources/values/strings.xml
```diff
@@ -514,6 +514,8 @@ Adjust volume by swiping vertically on the right side of the screen"</string>
             <string name="revanced_swipe_overlay_background_opacity_invalid_toast">Swipe opacity must be between 0-100</string>
             <string name="revanced_swipe_threshold_title">Swipe magnitude threshold</string>
             <string name="revanced_swipe_threshold_summary">The amount of threshold for swipe to occur</string>
+            <string name="revanced_swipe_volume_sensitivity_title">Volume swipe sensitivity</string>
+            <string name="revanced_swipe_volume_sensitivity_summary">Adjust volume change per swipe</string>
```
oSumAtrIX, line 518:
```suggestion
            <string name="revanced_swipe_volume_sensitivity_summary">The sensitivity of a swipe at which the volume should change</string>
```

extensions/youtube/src/main/java/app/revanced/extension/youtube/swipecontrols/controller/gesture/core/BaseGestureController.kt
```diff
@@ -24,6 +24,7 @@ abstract class BaseGestureController(
         controller.overlay,
         10,
         1,
+        controller.config.volumeSwipeSensitivity,
```
oSumAtrIX, line 27: @LisoUseInAIKyrios Do you know if it is correct to inject a specific config property into the interface implementation? This doesn't look like it is abstracted correctly here.

patches/src/main/resources/addresources/values/strings.xml
oSumAtrIX: So its the "stepSize". Probably a better name?

oSumAtrIX: Sensitivity strongly implies something else than step size. I was under the impression it changes how much you need to swipe to change the volume, but its in reality the step size. The summary can explain what happens such as "The amount of volume steps to change"

oSumAtrIX (approved):

PR #4576, opened 2025-03-09
feat(Spotify): Add `Spoof signature` patch

oSumAtrIX:


oSumAtrIX:
As the description suggests, it fixes functions broken when patched.

oSumAtrIX:
There is no detection mechanism or anything I saw that implies Spotify uses it to check for tampering, the signature affects the functionality of the app though.

oSumAtrIX:
The patch should apply to all late versions of the app

oSumAtrIX:
The official place is PlayStore, however it will only serve splits for Spotify.

emyaqin:
> > The official place is PlayStore, however it will only serve splits for Spotify.
> 
> Maybe there is a site like APKMIRROR that serve apk only. If PlayStore is the only one, then I'll try to merge app bundle with APKEditor then.

You can merge it easily with Antisplit-M though

oSumAtrIX:
Refer to the existing patches and their description

PR #4605, opened 2025-03-17
Added support for com.spotify.musix with Spotify patches

scrxpted7327:
Added (com.spotify.musix) to compatibility list for Spotify patches. com.spotify.musix is the package name for the cloned Spotify app from xManager.

### DIFF
--- a/patches/src/main/kotlin/app/revanced/patches/spotify/layout/theme/CustomThemePatch.kt
+++ b/patches/src/main/kotlin/app/revanced/patches/spotify/layout/theme/CustomThemePatch.kt
@@ -11,7 +11,10 @@ val customThemePatch = resourcePatch(
     name = "Custom theme",
     description = "Applies a custom theme.",
 ) {
-    compatibleWith("com.spotify.music")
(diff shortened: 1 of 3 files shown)


oSumAtrIX:
I'm unsure if we should add packages other than the official Spotify app. The patch is targeting Spotify. If we add xManager, then we would need to edit the code everytime we want to support another mod specifically. 

Instead, one could patch a spotify app using the change package name patch to change it back to the original one, then apply the patches they want along side the change package name patch again to change it back to what it was before.

@LisoUseInAIKyrios What do you say.

oSumAtrIX:
Xmanager Spotify doesn't benefit from the patch either as it ships the same fix after we released it

oSumAtrIX:
My personal gripe is more general. Adding individual packages isn't scalable properly. The approach I mentioned works for any Spotify package, the approach followed in this PR requires a code change every time. Not elegant. Regardless, given the patch isn't necessary for xmanager in the first place, we can close the PR. Thanks regardless!

PR #4651, opened 2025-03-26
Additional Premium Flags

xC3FFF0E:


sawankumarbundelkhandi:
finally @xC3FFF0E is here.

thank to both for you for patching spotify.

Nuckyz:
I looked at name, type, financial-product and catalogue when analyzing which attributes are needed. One of them enables the jam feature, which creating is server sided locked, the others didnt seem to do anything. Are we sure need those?

extensions/spotify/src/main/java/app/revanced/extension/spotify/misc/UnlockPremiumPatch.java
```diff
@@ -39,8 +39,8 @@ public final class UnlockPremiumPatch {
             // Enable Cross-Platform Spotify Car Thing.
             "can_use_superbird", true,
             // Disable the client-side subscription checks.
-            "com.spotify.madprops.delivered.by.ucs", "false",
-            "com.spotify.madprops.use.ucs.product.state", "false"
+            "com.spotify.madprops.delivered.by.ucs", false,
```
ILoveOpenSourceApplications, line 42: Shouldn't this be `true`?

Nuckyz: I am pretty sure it is enabled by default, will check in a second

ILoveOpenSourceApplications:
Also, will `"high-bitrate", true` provide better output than `High` in Spotify settings or no?

Brosssh:
It would enable the setting, but you can't play songs if you choose high quality. Just server side

0xrxL:
@xC3FFF0E Welcome legend 👋

Nuckyz:
@LisoUseInAIKyrios Make the custom theme patch disabled by default. We should not make the app amoled by default

0xrxL:
Only the package name patch is missing (to hide Play Store updates).

oSumAtrIX:
No, version code, but that already exists.

extensions/spotify/src/main/java/app/revanced/extension/spotify/misc/UnlockPremiumPatch.java
```diff
@@ -2,38 +2,57 @@
...
+        put("pick-and-shuffle", false);
+        // Disables shuffle-mode streaming-rule, which forces songs to be played shuffled
+        // and breaks the player when other patches are applied.
+        put("streaming-rules", "");
+        // Enables premium UI in settings and removes the premium button in the nav-bar.
+        put("nft-disabled", "1");
+        // Set the subscription name.
+        put("name", "Spotify Premium");
+        // Enable Cross-Platform Spotify Car Thing.
+        put("can_use_superbird", true);
```
oSumAtrIX, line 37: Is this confirmed to make any difference?

extensions/spotify/src/main/java/app/revanced/extension/spotify/misc/UnlockPremiumPatch.java
```diff
@@ -2,38 +2,57 @@
...
+        // Allows adding songs to queue and removes the smart shuffle mode restriction,
+        // allowing to pick any of the other modes.
+        put("pick-and-shuffle", false);
+        // Disables shuffle-mode streaming-rule, which forces songs to be played shuffled
+        // and breaks the player when other patches are applied.
+        put("streaming-rules", "");
+        // Enables premium UI in settings and removes the premium button in the nav-bar.
+        put("nft-disabled", "1");
+        // Set the subscription name.
+        put("name", "Spotify Premium");
```
oSumAtrIX, line 35: Where would that be visible in the UI?

oSumAtrIX (approved):

extensions/spotify/src/main/java/app/revanced/extension/spotify/misc/UnlockPremiumPatch.java
Nuckyz: not yet

extensions/spotify/src/main/java/app/revanced/extension/spotify/misc/UnlockPremiumPatch.java
0xrxL: Settings  --->  Account

Screenshot_20250327_020306_Spotify

oSumAtrIX:
Less than package name

extensions/spotify/src/main/java/app/revanced/extension/spotify/misc/UnlockPremiumPatch.java
oSumAtrIX: It already says Premium for me regardless of the flag override

extensions/spotify/src/main/java/app/revanced/extension/spotify/misc/UnlockPremiumPatch.java
oSumAtrIX: Waiting for brosssh?

extensions/spotify/src/main/java/app/revanced/extension/spotify/misc/UnlockPremiumPatch.java
Nuckyz: ya, unsure what it does but we should prob keep it for more consistency

extensions/spotify/src/main/java/app/revanced/extension/spotify/misc/UnlockPremiumPatch.java
Nuckyz: yep

extensions/spotify/src/main/java/app/revanced/extension/spotify/misc/UnlockPremiumPatch.java
0xrxL: > It already says Premium for me regardless of the flag override

In this case rely on Nuckyz explanation.

oSumAtrIX: > ya, unsure what it does but we should prob keep it for more consistency

I don't think we need it. Consistency would call for adding other fields which we haven't either. And if can_use_superbird also has no effect we'd get rid of it as well.

extensions/spotify/src/main/java/app/revanced/extension/spotify/misc/UnlockPremiumPatch.java
Brosssh: Tested it now. No changes both in mobile and android auto UI. Also, I still cannot choose to play a song with voice "play song bla bla", it will tell me that I need a premium subscription. Not sure where this check is done, maybe it's server sided

Brosssh (changes requested):
Remove superbird, doesn't change anything

JingMatrix:
> Also, I still cannot choose to play a song with voice "play song bla bla", it will tell me that I need a premium subscription. Not sure where this check is done, maybe it's server sided

For this, we may need to hook methods in `com.spotify.voiceassistants.playermodels.*`.
I tends to conjecture that it is not a server side feature.

oSumAtrIX:
Technically every patch does unless you mount. But because mounting is a thing, the signature patch isn't mandatory. Similar to how all patches don't depend on the spoof client patch in YT

ILoveOpenSourceApplications:
I'm getting ads in my home tab, should I open an issue?
Screenshot_20250327_194447_Spotify

ILoveOpenSourceApplications:
I'm getting ads in my home tab, should I open an issue?
Screenshot_20250327_194447_Spotify

JingMatrix:
@ILoveOpenSourceApplications Please specify your Spotify version. Did you use some systemwise AD blocker?  It not easy to reproduce your issue.

ILoveOpenSourceApplications:
> Please specify your Spotify version.

v9.0.28.630.

> Did you use some systemwise AD blocker? 

I have private dns set to `dns.adguard.com`.

> It not easy to reproduce your issue.

I can't either now.

Nuckyz:
> I'm getting ads in my home tab, should I open an issue? Screenshot_20250327_194447_Spotify
> 
> Edit: After closing and reopening a couple of times, the ad isn't showing up anymore. A rogue ad that bypassed the premium check, I guess?

No need to, the issue is known and we are working to avoid those ads from showing up

Brosssh (approved):

Purple-EyeZ:
There's this toast every time you open the app
Screenshot_20250328_180937_Spotify

Version: `9.0.30 642`

(I also tried newer versions but it's the same)

Jman-Github:
@Purple-EyeZ Being fixed here:

https://github.com/ReVanced/revanced-patches/pull/4688

Rai-08:
> There's this toast every time you open the app
> Screenshot_20250328_180937_Spotify
> 
> Version: `9.0.30 642`
> 
> (I also tried newer versions but it's the same)

In the order version i didn't have this problem but with the I'm also getting this every time I open the app

GSNinjaMusic (approved):

Issue #4682, opened 2025-03-28
bug: Spotify connect wont work as expected with alexa echo


andresmz:
### Bug description

When i try to use spotify connect with an amazon alexa, it will shuffle all the songs from my playlist, also will refuse to play a particular song and instead will play a random one.

### Error logs

```shell

```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] I have chosen an appropriate title.
- [x] All requested information has been provided properly.

Brosssh:
I have a similar behavior on an old phone when connecting to Android Auto. When asking for a specific song, it will play a station of that song.
I think everything that uses voice is done via `speakeasy` API, so it should be connected to this issue. 

What is weird is that on my actual phone i have a very different behavior. If I ask for a song, it will just tell me "You can't, buy premium first". 

Hope that we can get more feedback in order to clarify why this is happening. After we clarified we can fix it (maybe)

Brosssh:
A technical TLDR on how voice assistant works. This is being tested with Google Assistant, which is what most phones uses (via OK Google, Hey Google etc) and should also be used in Android Auto. Note that Gemini AI is NOT supported by Spotify as of now. Asking Gemini AI to play a song will open a browser page, telling you how to.

When a user does a voice search, 2 requests are executed:

- https://spclient.wg.spotify.com/speakeasy/v2/search?include=
This request send information about the device used for NLP processing, such as `google_assistant`, and a lot of other info, including the query used by the user (`play song bla bla`). It also contains a `parsed_query` object which already contains info regarding the track the user want to play

```
"parsed_query": {
        "intent": "PLAY",
        "uri": "spotify://track/0snQkGI5qnAmohLE7jTsTn"
}
```
After the request, the server will return the real action to execute based on user tier. If the user is a free/trial tier, it will return not the track, but instead a station with that track (aka random song, and at some point also the song you wanted, maybe). Core of this response if the field `action`

` "action": "spotify:nl:CAASEPCYM4aCE0wsl8kFqxRnucoaGzc3OjU6MHNuUWtHSTVxbkFtb2hMRTdqVHNUbiABMAPgAxPoA+n+0Y/eMvADIA=="`
 

- The second request after the one above is https://spclient.wg.spotify.com/speakeasy/v2/command/play:
The request take the `action` field from the previous response, send it to the server which then return `url` and `uri` of the object which need to be played (a song if premium, a station if free).


[...]

PR #4702, opened 2025-03-31
Bypass Voice Search restrictions

Brosssh:
When asking for a song with Google Assistant/Android Auto, if the user is free tier a station/radio containing the requested song is played. This PR will fix this issue, allowing patched app to play requested song/artist.

This should also partially solve #4682, although I don't have Alexa so i can't test it myself. I tested it with both Android Auto and Ok Google and it's working fine. Asking for a song will play that song. Asking for an author will play Spotify choices for the author (which is the same behavior as real premium users).


patches/src/main/kotlin/app/revanced/patches/spotify/misc/voicesearch/BypassVoiceSearchRestrictions.kt
```diff
@@ -0,0 +1,38 @@
...
+            val registerUrl = getInstruction<FiveRegisterInstruction>(insertIndex).registerC
+            val registerUri = getInstruction<FiveRegisterInstruction>(insertIndex + 2).registerD
+
+            addInstructions(
+                insertIndex,
+            """
+                invoke-static { v$registerUrl }, $EXTENSION_CLASS_DESCRIPTOR->removeStationString(Ljava/lang/String;)Ljava/lang/String;
+                move-result-object v$registerUrl
+                invoke-static { v$registerUri }, $EXTENSION_CLASS_DESCRIPTOR->removeStationString(Ljava/lang/String;)Ljava/lang/String;
+                move-result-object v$registerUri
```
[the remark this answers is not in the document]
Brosssh: Not a typo, one is ur**i** (letter i) other is ur**l** (letter l). Probably not the best name but that's what spotify uses. (Unless I'm not seeing a typo)

Brosssh:
Maybe the patch description can be adjusted, my English is pretty bad

Brosssh:
Found a way to remove forced shuffle as well. I'll add it directly to this PR since it's another forced restriction via voice search

Nuckyz:
Why can't this be part of unlock premium?

Brosssh:
I would keep the premium patch as easy and as clean as possible. This is a pretty niche feature and it should not impact the behavior of the normal premium patch, in my opinion. If something in this patch breaks, it might not necessarily break something in the premium patch, but merging them together could impact the premium patch.

But if other people want to merge then together we can, ofc. Just want to find the best solution

oSumAtrIX:
Try catch blocks can be used, however a patch shouldn't be designed with breaking in mind. If you fear it breaking, then it should be rewritten in a way which avoids that. Since this feature is a premium one, merging it to the premium unlock patch does make sense for simplicity, if people use the premium patch, they will want this one and vice versa

Brosssh:
I don't necessarily agree. I would prefer a patch to break hard rather than silently inside a try catch, so we know it's broken and needs to be fixed. And having it separate is useful for this case, so it can be easily toggled off without issues (giving its niche utility).

Obviously the patch is not designed to break easily (in fact it's probably not going to break anytime soon), it's patching the json reader method which reads the response, so unless the response changes a lot there will be no issue. 

I of course value your feedbacks and experience so I'll merge the patches. But I think it's always useful to see other points of view, so that's why I wanted to raise my point of view 🙂

JingMatrix:
@Brosssh  In my device, the google voice assistance can play music without problem (without this patch applied), but I am still curious about this patch.
Using Xposed hooking, I recoreded the call of `fromJson` using `toString()`:
1. the parameter is `JsonReader(buffer(t46(buffer(p.o2q@8f454d9)))) `
2. the result is `Context{uri=spotify:playlist:37i9dQZF1DZ06evO1SVXaM, url=context://spotify:playlist:37i9dQZF1DZ06evO1SVXaM, metadata={autoplay_candidate=true}, pages=Optional.absent(), restrictions=Optional.absent()}`

Could you please explain which parameters in the `com.spotify.player.model.Context` object is changed by current patch?

Brosssh:
You can read my technical tldr here. https://github.com/ReVanced/revanced-patches/issues/4682#issuecomment-2763362949

In short, I am modifying both uri an url, removing the `:station` text. That text is added when a free account requests a specific song. Can you choose a specific song via voice assistant?

JingMatrix:
Okay, after testing with two songs, the voice assistance refuses to paly a specific asked track.
And you are right, I have now found the `station` string in the Context object: `Context{uri=spotify:station:track:4XNrMwGx1SqP01sqkGTDmo, url=context://spotify:station:track:4XNrMwGx1SqP01sqkGTDmo, metadata={autoplay_candidate=true}, pages=Optional.absent(), restrictions=Optional.absent()}`.

Thanks for sharing you knowledge.
I will test playlist and tell you the results later.

JingMatrix:
After removing the `station:` part, the requested music tracks are played as expected.

You were right, the playlist are shuffled. Actually, it is known from the UI that the `shuffle play` mode is turned on.

andresmz:
Is there a way to disable on UI that auto shuffle mode on when you connect to a voice device?

oSumAtrIX:
> I don't necessarily agree. I would prefer a patch to break hard rather than silently inside a try catch, so we know it's broken and needs to be fixed. And having it separate is useful for this case, so it can be easily toggled off without issues (giving its niche utility).
> 
> Obviously the patch is not designed to break easily (in fact it's probably not going to break anytime soon), it's patching the json reader method which reads the response, so unless the response changes a lot there will be no issue.
> 
> I of course value your feedbacks and experience so I'll merge the patches. But I think it's always useful to see other points of view, so that's why I wanted to share my point of view 🙂

Patches can use the logger to log anything useful. Catching an exception inside a patch means the patch is considered successful. You can save the exception, then at the end of the patch rethrow it, or dismiss it and show a log, the possibilities are endless. However what's clear is that this patch is what belongs to the Premium unlock patch. Given that, if this patch fails, the premium unlock should be considered failed, and not successful. This is desired behavior. This doesn't mean that you can't design it so that it breaks the other patches the premium unlock patch makes. Catching the exception, letting the other patches run then rethrowing is for example valid.

Brosssh:
@andresmz @JingMatrix not sure about what UI you are talking about.. can you maybe show a screenshot of it?

JingMatrix:
share_8395724755558067077

Brosssh:
With this PR, asking for a playlist/album will play that NOT in shuffle. If you want to shuffle, ask for the playlist/album in shuffle order. This behavior is the same in real premium. 

You can always shuffle/unshuffle a playlist via the shuffle button, at least I had no issue doing this.

Brosssh:
There is no `license` field both in free and premium response. Probably deprecated, or to be implemented.

As of now, this PR should already be emulating premium behavior fully.

patches/src/main/kotlin/app/revanced/patches/spotify/misc/UnlockPremiumPatch.kt
```diff
@@ -45,5 +50,38 @@ val unlockPremiumPatch = bytecodePatch(
...
+                """
+                    invoke-static { v$registerUrl }, $extensionMethodDescriptor
+                    move-result-object v$registerUrl
+                    invoke-static { v$registerUri }, $extensionMethodDescriptor
+                    move-result-object v$registerUri
+                """
+            )
+        }
+
+        // Removed forced shuffle when asking for an album/playlist via Voice assistant.
```
oSumAtrIX, line 74:
```suggestion
        // Disable forced shuffle when asking for an album/playlist via Voice assistant.
```

extensions/spotify/src/main/java/app/revanced/extension/spotify/premium/UnlockPremiumPatch.java
```diff
@@ -1,4 +1,4 @@
-package app.revanced.extension.spotify.misc;
+package app.revanced.extension.spotify.premium;
```
oSumAtrIX, line 1: The package should be the same as in the patch for consistency. So far premium patches have been under misc and not a dedicated package.

oSumAtrIX (approved):
Is "Voice assistant" a brand name? Otherwise V needs to be lowercase

andresmz:
> Does this "only" cover voice assistent or the general Spotify Connect connectivity to all device types. It is behaving similarily from a user perspective and what issue #4682 also mentions?

We need to try this patch first, but they dont have an Alexa device to test it right now, im trying to patch it but have not been successfull yet

Brosssh:
> Does this "only" cover voice assistent or the general Spotify Connect connectivity to all device types. It is behaving similarily from a user perspective and what issue #4682 also mentions?

This PR should fix the mentioned issue. I don't have Alexa myself so I cannot test it 😄 , but I am assuming that all voice-related devices use the same API (`speakeasy`), so they will all get fixed by this. 

I tested Android Auto and Google assistant, so I can speak only for them. But I'm confident it will fix others (if not all).

andresmz:
> > Does this "only" cover voice assistent or the general Spotify Connect connectivity to all device types. It is behaving similarily from a user perspective and what issue #4682 also mentions?
> 
> This PR should fix the mentioned issue. I don't have Alexa myself so I cannot test it 😄 , but I am assuming that all voice-related devices use the same API (`speakeasy`), so they will all get fixed by this.
> 
> I tested Android Auto and Google assistant, so I can speak only for them. But I'm confident it will fix others (if not all).

I just tried with this patch patches-5.18.1-dev.1.rvp and with the latest artifact (https://github.com/ReVanced/revanced-patches/actions/runs/14214441513/artifacts/2865573771) (java -jar revanced-cli-5.0.1-dev.1-all.jar  patch  -p patches-5.19.0-dev.1.rvp -o test.apk "com.spotify.music v9.0.30.642_antisplit.apk") but it didnt work on amazon echo, still plays a random song from a playlist, also after a few songs it will force UI to random too, even if i disable that button it will still plays a random song.

Brosssh:
Given that this fix doesn't work on Alexa (according to @andresmz), I changed the title. I think it can be merged as a specific fix for Google Assistant and then someone else can create a specific fix for Alexa. Unfortunately I cannot help with that, I don't have the device.

Brosssh:
I have nothing else to do regarding this PR btw, not sure if you are waiting for something else from my side or just didn't have time to look at it yet.

ILoveOpenSourceApplications:
Since #4682 specifically asks for a solution for Amazon Alexa Echo, I think the issue should be unlinked.

patches/src/main/kotlin/app/revanced/patches/spotify/misc/UnlockPremiumPatch.kt
```diff
@@ -62,47 +66,82 @@ val unlockPremiumPatch = bytecodePatch(
             method.replaceInstruction(moveIsEnabledIndex, "const/4 v$isUpsellEnabledRegister, 0")
         }
 
+        // Enable choosing a specific song/artist via Google Assistant.
+        contextFromJsonFingerprint.method.apply {
+            val insertIndex = contextFromJsonFingerprint.patternMatch!!.startIndex
+            val registerUrl = getInstruction<FiveRegisterInstruction>(insertIndex).registerC
+            val registerUri = getInstruction<FiveRegisterInstruction>(insertIndex + 2).registerD
```
oSumAtrIX, line 73: Why is there a Url and Uri?

patches/src/main/kotlin/app/revanced/patches/spotify/misc/UnlockPremiumPatch.kt
```diff
@@ -62,47 +66,82 @@ val unlockPremiumPatch = bytecodePatch(
...
+                    move-result-object v$registerUri
+                """
+            )
+        }
+
+        // Disable forced shuffle when asking for an album/playlist via Google Assistant.
+        readPlayerOptionOverridesFingerprint.method.apply {
+            val shufflingContextCallIndex = indexOfFirstInstructionOrThrow {
+                getReference<MethodReference>()?.name == "shufflingContext"
+            }
```
oSumAtrIX, line 93: IIRC theres a helper method to find the first method reference? I forgot though @LisoUseInAIKyrios was that added?

patches/src/main/kotlin/app/revanced/patches/spotify/misc/UnlockPremiumPatch.kt
```diff
@@ -62,47 +66,82 @@ val unlockPremiumPatch = bytecodePatch(
...
-        with(protobufListRemoveFingerprint.method) {
-            val invokeThrowUnmodifiableIndex = indexOfFirstInstructionOrThrow {
-                val reference = getReference<MethodReference>()
-                opcode == Opcode.INVOKE_VIRTUAL &&
-                        reference?.returnType == "V" && reference.parameterTypes.isEmpty()
+            val protobufListRemoveFingerprint = fingerprint {
+                custom { method, classDef ->
+                    method.name == "remove" && classDef.type == protobufListClassName
+                }
             }
```
oSumAtrIX, line 117: It is possible to match a fingerprint to a method via "myFingerprint.match(classDef)". Use that to match against a class. What you have now matches against all classes.

patches/src/main/kotlin/app/revanced/patches/spotify/misc/UnlockPremiumPatch.kt
```diff
@@ -62,47 +66,82 @@ val unlockPremiumPatch = bytecodePatch(
...
-        }
+            // Need to allow mutation of the list so the home ads sections can be removed.
+            // Protobuffer list has an 'isMutable' boolean parameter that sets the mutability.
+            // Forcing that always on breaks unrelated code in strange ways.
+            // Instead, remove the method call that checks if the list is unmodifiable.
+            with(protobufListRemoveFingerprint.method) {
+                val invokeThrowUnmodifiableIndex = indexOfFirstInstructionOrThrow {
+                    val reference = getReference<MethodReference>()
+                    opcode == Opcode.INVOKE_VIRTUAL &&
+                            reference?.returnType == "V" && reference.parameterTypes.isEmpty()
```
oSumAtrIX, line 127: The opcode check may not be necessary:

```suggestion
                    val reference = getReference<MethodReference>()
                    	?: return@indexOfFirstInstructionOrThrow null
                    	
                    reference.returnType == "V" && reference.parameterTypes.isEmpty()
```

if the opcode is required then keep it and use the null check operator

patches/src/main/kotlin/app/revanced/patches/spotify/misc/UnlockPremiumPatch.kt
```diff
@@ -62,47 +66,82 @@ val unlockPremiumPatch = bytecodePatch(
...
+                }
+
+                // Remove the method call that throws an exception if the list is not mutable.
+                removeInstruction(invokeThrowUnmodifiableIndex)
```
oSumAtrIX, line 131: you can use this syntax indexOf...().let(::removeInstruction)

oSumAtrIX (commented):
LGTM, can be merged once the comments are resolved.

patches/src/main/kotlin/app/revanced/patches/spotify/misc/UnlockPremiumPatch.kt
oSumAtrIX: Oh no thats not what I meant. Instead of `classDef.type == protobufListClassName`, simply use 

```kt
fingerprint {
    custom { method, classDef -> method.name == "remove" }
}.match(protobufListClassDef)
```

Nuckyz: What happened here? Why does this pr overwrite so many changes from my other pr which was already merged. Seems like it was rebased incorrectly.

oSumAtrIX: In this case use first find the class in this@execute.classes or use the findClassBy function.

Nuckyz: The class is found by looking at another method which references it. I dont think those functions will help.

oSumAtrIX: No i mean, you already have protobufListClassName so just do 

fingerprint.match(classBy { it.name == protobufListClassName }.originalClassDef)

Nuckyz: I see. I will do it in my other pr once this is merged.

Nuckyz: > I think the changes crept in because the code got formatted, not sure though. I reverted all unrelated changes.

I will take the suggestions and refactor from my other PR.

extensions/spotify/src/main/java/app/revanced/extension/spotify/misc/UnlockPremiumPatch.java
```diff
@@ -105,4 +112,4 @@ public static void removeHomeSections(List<Section> sections) {
             Logger.printException(() -> "Remove home sections failure", ex);
         }
     }
-}
+}
```
Nuckyz, line 115:
```suggestion
}

```

patches/src/main/kotlin/app/revanced/patches/spotify/misc/Fingerprints.kt
```diff
@@ -34,4 +55,4 @@ internal val protobufListsFingerprint = fingerprint {
 internal val homeStructureFingerprint = fingerprint {
     opcodes(Opcode.IGET_OBJECT, Opcode.RETURN_OBJECT)
     custom { _, classDef -> classDef.endsWith("homeapi/proto/HomeStructure;") }
-}
+}
```
Nuckyz, line 58:
```suggestion
}

```

Nuckyz (changes requested):

patches/src/main/kotlin/app/revanced/patches/spotify/misc/UnlockPremiumPatch.kt
```diff
@@ -62,6 +66,40 @@ val unlockPremiumPatch = bytecodePatch(
             method.replaceInstruction(moveIsEnabledIndex, "const/4 v$isUpsellEnabledRegister, 0")
         }
 
+        // Enable choosing a specific song/artist via Google Assistant.
```
Nuckyz, line 69: We should move the patches added by this pr before the upsells one. The functionality patches should come before the visual fixes patches.

Nuckyz (changes requested):

PR #4765, opened 2025-04-11
feat(Spotify): Add `Check environment` patch

oSumAtrIX:


oSumAtrIX:
Screenshot_20250412-144830

oSumAtrIX:
In addition I have moved the context hook into finalize. This way the context is hooked at last and thus inserted before any other patch depending on the shared extension patch. This solves the temporal fix done in the gmscore support patch.

Issue #4675, opened 2025-03-28
bug(Spotify): setting playback device to remote is paywalled


zdimension:
### Bug description

On 8.x, I used to be able to open Spotify on my phone, click on the "Devices" button and click on my computer (on which Spotify desktop is running) to be able to start playing music on it, while controlling from my phone. Now, it's greyed out:

![Image](https://github.com/user-attachments/assets/a97a423a-5ca5-4b93-8748-bb80dfb7b95f)

The workaround is to start a song on my phone, then go to my computer (or use a remote desktop app) and set the playback device to my computer, and then I can keep controlling the playing song from my phone. But it's unfortunate that I can't start the playback on my computer from my phone anymore.

### Error logs

```shell

```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] I have chosen an appropriate title.
- [x] All requested information has been provided properly.

zdimension:
All other features work, yep

Nuckyz:
@LisoUseInAIKyrios you can close this issue, no other issue reported it.

Brosssh:
This is getting more frequent. I have it on both of my phones now and I saw other people reporting it. I don't know if it's related to the app version, or rather controlled by an experiment.

Anyway I'm verifying to see if it can be bypassed.

PR #4782, opened 2025-04-13
fix(Spotify - Unlock Spotify Premium): Remove premium restriction on Spotify Connect 

Brosssh:
For some people, including myself, Spotify Connect is restricted behind Premium subscription. This PR simply removes this restriction. This would also close https://github.com/ReVanced/revanced-patches/issues/4675 (although is already closed since no other people reported it).

Screenshot_2025-04-13-20-42-18-12_0438eb925998df20b3482ec25499d226 (1)

### DIFF
--- a/extensions/spotify/src/main/java/app/revanced/extension/spotify/misc/UnlockPremiumPatch.java
+++ b/extensions/spotify/src/main/java/app/revanced/extension/spotify/misc/UnlockPremiumPatch.java
@@ -80,14 +80,14 @@ public final class UnlockPremiumPatch {
             new OverrideAttribute("streaming-rules", ""),
             // Enables premium UI in settings and removes the premium button in the nav-bar.
             new OverrideAttribute("nft-disabled", "1"),
+            // Removes Spotify Connect as a premium only feature.
+            new OverrideAttribute("type", "premium"),
             // Enable Spotify Car Thing hardware device.
             // Device is discontinued and no longer works with the latest releases,
             // but it might still work with older app targets.
             new OverrideAttribute("can_use_superbird", TRUE, false),
             // Removes the premium button in the nav-bar for tablet users.
-            new OverrideAttribute("tablet-free", FALSE, false),
-            // Removes Spotify Connect as a premium only feature.
-            new OverrideAttribute("type", "premium")
+            new OverrideAttribute("tablet-free", FALSE, false)
     );
 
     private static final List<Integer> REMOVED_HOME_SECTIONS = List.of(
--- a/extensions/spotify/src/main/java/app/revanced/extension/spotify/misc/UnlockPremiumPatch.java
+++ b/extensions/spotify/src/main/java/app/revanced/extension/spotify/misc/UnlockPremiumPatch.java
@@ -80,7 +80,8 @@ public final class UnlockPremiumPatch {
             new OverrideAttribute("streaming-rules", ""),
             // Enables premium UI in settings and removes the premium button in the nav-bar.
             new OverrideAttribute("nft-disabled", "1"),
-            // Removes Spotify Connect as a premium only feature.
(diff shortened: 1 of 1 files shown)


extensions/spotify/src/main/java/app/revanced/extension/spotify/misc/UnlockPremiumPatch.java
```diff
@@ -85,7 +85,9 @@ private static class OverrideAttribute {
             // but it might still work with older app targets.
             new OverrideAttribute("can_use_superbird", TRUE, false),
             // Removes the premium button in the nav-bar for tablet users.
-            new OverrideAttribute("tablet-free", FALSE, false)
+            new OverrideAttribute("tablet-free", FALSE, false),
+            // Removes Spotify Connect as a premium only feature.
```
Nuckyz, line 89:
```suggestion
            // Enable Spotify Connect and disable other premium related UI, like buying premium.
```

Nuckyz (changes requested):

Brosssh:
I just double checked and it's present in `8.6.98.900` aswell. I moved it before optional attributes.

extensions/spotify/src/main/java/app/revanced/extension/spotify/misc/UnlockPremiumPatch.java
Brosssh: > and disable other premium related UI, like buying premium.
Is this changing the UI?

Nuckyz: Yes, it also makes the download songs button disappear if you offline is set to false, which is what free accounts have

Brosssh: Can't we just override `offline` as well then?

Nuckyz: Offline means downloading, that's server side locked

drobotk:
Worth noting that this doesn't enable using standalone Spotify Connect clients. That feature still requires premium as it's probably server-sided.

extensions/spotify/src/main/java/app/revanced/extension/spotify/misc/UnlockPremiumPatch.java
Brosssh: Obviously it won't enable downloading, but it will bring back the download button as of today. If there are no other effects in enabling `offline`, I see no issues.

Nuckyz: No, it will enable the download button and allow the user to think they can download, but it will just error in the request. Don't change that flag.

Nuckyz: It's fine to change the type flag. Just worth to mention in the description it removes some premium related UI.

Brosssh: My brain just can't process why enabling `type` `premium` would make the download button disappear... which is a premium feature.

drobotk: On free accounts, the download button is there to upsell Spotify Premium. If we now spoof type=premium, and also disable downloading (offline=false), the download button disappears

Brosssh:
I'm not a big fan of the fact that the download button disappears. If needed, we can set `offline` flag to true and restore free account behaviour on the download button press with a bytecode patch (so we force it to show "you have to buy premium", as of now).

oSumAtrIX:
A separate PR can be made to achieve this alltogether.

Issue #4676, opened 2025-03-28
feat: Prime Video Hide Ads


realdcre:
### Feature description

A patch for Amazon Prime Video, that allows ads to be blocked, simmilar to YouTube.

### Motivation

Since Prime Video has added ads in its Standard plan, the App has become annoying to use. The benefits are clear: No Ads, for all people.

### Acknowledgements

- [x] I have checked all open and closed feature requests and this is not a duplicate
- [x] I have chosen an appropriate title.
- [x] All requested information has been provided properly.

the-wondersmith:
Apologies for resurrecting a closed issue - any chance these patches will work / can be enabled for the Chromecast version of the app (`com.amazon.amazonvideo.livingroom.apk`)?

hoo-dles:
> [@hoo-dles](https://github.com/hoo-dles) / [@LisoUseInAIKyrios](https://github.com/LisoUseInAIKyrios)
> 
> Apologies for resurrecting a closed issue - any chance these patches will work / can be enabled for the Chromecast version of the app (`com.amazon.amazonvideo.livingroom.apk`)?

Open a new ticket and I'll take a look

PR #4824, opened 2025-04-18
feat(Prime Video): Initial commit of Skip Ads patch

hoo-dles:
This is a first pass of a patch for Amazon Prime Video to skip ads. Ad breaks seem to be baked into the incoming stream, so the best we can do is force the video player to seek over them.

I could use some help with more extensive testing since the Prime Video app has like 5 different video player implementations and I'm not sure when (if?) the ones that I didn't patch are used.

### DIFF
--- a/extensions/primevideo/src/main/java/app/revanced/extension/primevideo/ads/SkipAdsPatch.java
+++ b/extensions/primevideo/src/main/java/app/revanced/extension/primevideo/ads/SkipAdsPatch.java
@@ -1,0 +1,25 @@
+package app.revanced.extension.primevideo.ads;
+
+import com.amazon.avod.fsm.SimpleTrigger;
+import com.amazon.avod.media.ads.AdBreak;
+import com.amazon.avod.media.ads.internal.state.AdBreakTrigger;
+import com.amazon.avod.media.ads.internal.state.AdEnabledPlayerTriggerType;
+import com.amazon.avod.media.playback.VideoPlayer;
+import com.amazon.avod.media.ads.internal.state.ServerInsertedAdBreakState;
+
+@SuppressWarnings("unused")
+public final class SkipAdsPatch {
+    public static void ServerInsertedAdBreakState_enter(ServerInsertedAdBreakState state, AdBreakTrigger trigger, VideoPlayer player) {
+        AdBreak adBreak = trigger.getBreak();
+
+        if (trigger.getSeekStartPosition() != null)
+            // if scrubbing over ad, seek straight to it
+            player.seekTo(trigger.getSeekTarget().getTotalMilliseconds());
+        else
+            // if naturally entering ad, seek to end of ad break
+            player.seekTo(player.getCurrentPosition() + adBreak.getDurationExcludingAux().getTotalMilliseconds());
+
+        // send "end of ads" trigger to state machine so everything doesn't get whacky
+        state.doTrigger(new SimpleTrigger(AdEnabledPlayerTriggerType.NO_MORE_ADS_SKIP_TRANSITION));
+    }
+}
--- a/extensions/primevideo/stub/build.gradle.kts
+++ b/extensions/primevideo/stub/build.gradle.kts
@@ -1,0 +1,17 @@
+plugins {
+    id(libs.plugins.android.library.get().pluginId)
+}
+
+android {
+    namespace = "app.revanced.extension"
+    compileSdk = 34
+
+    defaultConfig {
+        minSdk = 26
+    }
+
+    compileOptions {
+        sourceCompatibility = JavaVersion.VERSION_17
+        targetCompatibility = JavaVersion.VERSION_17
+    }
+}
--- a/patches/src/main/kotlin/app/revanced/patches/primevideo/ads/Fingerprints.kt
+++ b/patches/src/main/kotlin/app/revanced/patches/primevideo/ads/Fingerprints.kt
@@ -1,0 +1,34 @@
+package app.revanced.patches.primevideo.ads
+
+import app.revanced.patcher.fingerprint
+import com.android.tools.smali.dexlib2.AccessFlags
+import com.android.tools.smali.dexlib2.Opcode
+
+internal val enterServerInsertedAdBreakStateFingerprint = fingerprint {
+    accessFlags(AccessFlags.PUBLIC)
+    parameters("Lcom/amazon/avod/fsm/Trigger;")
+    returns("V")
+    opcodes(
+        Opcode.INVOKE_VIRTUAL,
+        Opcode.MOVE_RESULT_OBJECT,
+        Opcode.CONST_4,
+        Opcode.CONST_4
+    )
+    custom { method, classDef ->
+        classDef.type == "Lcom/amazon/avod/media/ads/internal/state/ServerInsertedAdBreakState;" &&
+                method.name == "enter"
+    }
+}
+
+internal val doTriggerFingerprint = fingerprint {
+    accessFlags(AccessFlags.PROTECTED)
+    returns("V")
+    opcodes(
+        Opcode.IGET_OBJECT,
+        Opcode.INVOKE_INTERFACE,
+        Opcode.RETURN_VOID
+    )
+    custom { method, classDef ->
+        classDef.type == "Lcom/amazon/avod/fsm/StateBase;" && method.name == "doTrigger"
+    }
+}
--- a/patches/src/main/kotlin/app/revanced/patches/primevideo/ads/SkipAdsPatch.kt
+++ b/patches/src/main/kotlin/app/revanced/patches/primevideo/ads/SkipAdsPatch.kt
@@ -1,0 +1,45 @@
+package app.revanced.patches.primevideo.ads
+
+import app.revanced.patcher.extensions.InstructionExtensions.addInstructions
+import app.revanced.patcher.extensions.InstructionExtensions.getInstruction
+import app.revanced.patcher.patch.bytecodePatch
+import app.revanced.patches.shared.misc.extension.sharedExtensionPatch
+import com.android.tools.smali.dexlib2.AccessFlags
+import com.android.tools.smali.dexlib2.iface.instruction.OneRegisterInstruction
+
+@Suppress("unused")
+val skipAdsPatch = bytecodePatch(
+    name = "Skip ads",
+    description = "Automatically skip over pre-baked ads in video streams.",
+) {
+    compatibleWith("com.amazon.avod.thirdpartyclient")
+
+    dependsOn(sharedExtensionPatch("primevideo"))
+
+    // We're going to skip all the logic in ServerInsertedAdBreakState.enter(), which plays all the ad clips in this
+    // ad break. Instead, we'll force the video player to seek over the entire break and reset the state machine.
+    execute {
+        // Force doTrigger() access to public so we can call it from our extension
+        doTriggerFingerprint.method.accessFlags = AccessFlags.PUBLIC.value;
+
+        val getPlayerIndex = enterServerInsertedAdBreakStateFingerprint.patternMatch!!.startIndex
+        enterServerInsertedAdBreakStateFingerprint.method.apply {
+            // Get register that stores VideoPlayer:
+            //  invoke-virtual ->getPrimaryPlayer()
+            //  move-result-object {vx}
+            val playerRegister = getInstruction<OneRegisterInstruction>(getPlayerIndex + 1).registerA
+
+            // Reuse the params from the original method
+            //  p0 = ServerInsertedAdBreakState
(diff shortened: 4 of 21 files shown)


hoo-dles:
Related to feature request: #4676

extensions/primevideo/stub/build.gradle.kts
```diff
@@ -0,0 +1,17 @@
...
+    namespace = "app.revanced.extension"
+    compileSdk = 34
+
+    defaultConfig {
+        minSdk = 26
+    }
+
+    compileOptions {
+        sourceCompatibility = JavaVersion.VERSION_17
+        targetCompatibility = JavaVersion.VERSION_17
```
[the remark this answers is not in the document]
kitadai31: `VERSION_17` should works with minSdk 21

It is d8's job to compile the Java 17 .class files into dex bytecodes, but the latest d8 should support Java 17 bytecodes.

If this is correct, the `compileOptions` can be removed as `VERSION_17` is the default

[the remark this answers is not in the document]
kitadai31: I understand.

When I tried record class recently, it worked with Android 6.0
So I had thought Java 17 should work.

But it was a standalone app, not ReVanced extensions...
I have never tested in ReVanced Extensions.
Maybe using records in ReVanced Extensions causes a problem with merging the d8's desugar classes?

areteruhiro:
Hello, I am developing an Xposed module that skips ads on Prime Video. I hope my module can be of help to you. Also, I don't plan to publish it on the Xposed Modules Repository in order to avoid being countered by Prime.


Also, the only app that can be installed on most Android TVs is "com.amazon.amazonvideo.livingroom". (I understand that there are shared components, but personally, I believe it's necessary to create a separate patch specifically for this Android TV version of the app.)

patches/src/main/kotlin/app/revanced/patches/primevideo/ads/SkipAdsPatch.kt
```diff
@@ -0,0 +1,45 @@
...
+import app.revanced.patcher.extensions.InstructionExtensions.getInstruction
+import app.revanced.patcher.patch.bytecodePatch
+import app.revanced.patches.shared.misc.extension.sharedExtensionPatch
+import com.android.tools.smali.dexlib2.AccessFlags
+import com.android.tools.smali.dexlib2.iface.instruction.OneRegisterInstruction
+
+@Suppress("unused")
+val skipAdsPatch = bytecodePatch(
+    name = "Skip ads",
+    description = "Automatically skip video stream ads.",
```
oSumAtrIX, line 13:
```suggestion
    description = "Automatically skips video stream ads.",
```

patches/src/main/kotlin/app/revanced/patches/primevideo/ads/SkipAdsPatch.kt
```diff
@@ -0,0 +1,45 @@
...
+) {
+    compatibleWith("com.amazon.avod.thirdpartyclient"("3.0.403.257"))
+
+    dependsOn(sharedExtensionPatch("primevideo"))
+
+    // We're going to skip all the logic in ServerInsertedAdBreakState.enter(), which plays all the ad clips in this
+    // ad break. Instead, we'll force the video player to seek over the entire break and reset the state machine.
```
oSumAtrIX, line 20:
```suggestion
    // Skip all the logic in ServerInsertedAdBreakState.enter(), which plays all the ad clips in this
    // ad break. Instead, force the video player to seek over the entire break and reset the state machine.
```

extensions/primevideo/stub/build.gradle.kts
```diff
@@ -0,0 +1,17 @@
+plugins {
+    id(libs.plugins.android.library.get().pluginId)
+}
+
+android {
+    namespace = "app.revanced.extension"
```
oSumAtrIX, line 6: Isn't this set already by the patches Gradle plugin?

extensions/primevideo/stub/build.gradle.kts
```diff
@@ -0,0 +1,17 @@
...
+    compileSdk = 34
+
+    defaultConfig {
+        minSdk = 21
+    }
```
oSumAtrIX, line 11: Isn't this also set already by the patches Gradle plugin?

extensions/primevideo/src/main/java/app/revanced/extension/primevideo/ads/SkipAdsPatch.java
```diff
@@ -0,0 +1,25 @@
...
+import com.amazon.avod.fsm.SimpleTrigger;
+import com.amazon.avod.media.ads.AdBreak;
+import com.amazon.avod.media.ads.internal.state.AdBreakTrigger;
+import com.amazon.avod.media.ads.internal.state.AdEnabledPlayerTriggerType;
+import com.amazon.avod.media.playback.VideoPlayer;
+import com.amazon.avod.media.ads.internal.state.ServerInsertedAdBreakState;
+
+@SuppressWarnings("unused")
+public final class SkipAdsPatch {
+    public static void ServerInsertedAdBreakState_enter(ServerInsertedAdBreakState state, AdBreakTrigger trigger, VideoPlayer player) {
```
oSumAtrIX, line 12: I don't think the naming convention we use for extensions is respected here. Please match the style. Something like "enterServerInsertedAdBreakState"

extensions/primevideo/src/main/java/app/revanced/extension/primevideo/ads/SkipAdsPatch.java
```diff
@@ -0,0 +1,25 @@
...
+        if (trigger.getSeekStartPosition() != null)
+            // if scrubbing over ad, seek straight to it
+            player.seekTo(trigger.getSeekTarget().getTotalMilliseconds());
+        else
+            // if naturally entering ad, seek to end of ad break
+            player.seekTo(player.getCurrentPosition() + adBreak.getDurationExcludingAux().getTotalMilliseconds());
+
+        // send "end of ads" trigger to state machine so everything doesn't get whacky
+        state.doTrigger(new SimpleTrigger(AdEnabledPlayerTriggerType.NO_MORE_ADS_SKIP_TRANSITION));
+    }
```
oSumAtrIX, line 24:
```suggestion
        if (trigger.getSeekStartPosition() != null)
            // If scrubbing over an ad, seek straight to it.
            player.seekTo(trigger.getSeekTarget().getTotalMilliseconds());
        else
            // If naturally entering an ad, seek to end the ad break.
            player.seekTo(player.getCurrentPosition() + adBreak.getDurationExcludingAux().getTotalMilliseconds());

        // Send "end of ads" trigger to state machine so everything doesn't get whacky.
        state.doTrigger(new SimpleTrigger(AdEnabledPlayerTriggerType.NO_MORE_ADS_SKIP_TRANSITION));
    }
```

patches/src/main/kotlin/app/revanced/patches/primevideo/ads/SkipAdsPatch.kt
```diff
@@ -0,0 +1,45 @@
...
+            //  move-result-object {vx}
+            val playerRegister = getInstruction<OneRegisterInstruction>(getPlayerIndex + 1).registerA
+
+            // Reuse the params from the original method:
+            //  p0 = ServerInsertedAdBreakState
+            //  p1 = AdBreakTrigger
+            addInstructions(
+                getPlayerIndex + 2,
+                """
+                    invoke-static {p0, p1, v${playerRegister}}, Lapp/revanced/extension/primevideo/ads/SkipAdsPatch;->ServerInsertedAdBreakState_enter(Lcom/amazon/avod/media/ads/internal/state/ServerInsertedAdBreakState;Lcom/amazon/avod/media/ads/internal/state/AdBreakTrigger;Lcom/amazon/avod/media/playback/VideoPlayer;)V
```
oSumAtrIX, line 38:
```suggestion
                    invoke-static { p0, p1, v$playerRegister }, Lapp/revanced/extension/primevideo/ads/SkipAdsPatch;->ServerInsertedAdBreakState_enter(Lcom/amazon/avod/media/ads/internal/state/ServerInsertedAdBreakState;Lcom/amazon/avod/media/ads/internal/state/AdBreakTrigger;Lcom/amazon/avod/media/playback/VideoPlayer;)V
```

patches/src/main/kotlin/app/revanced/patches/primevideo/ads/SkipAdsPatch.kt
```diff
@@ -0,0 +1,45 @@
...
+            //  invoke-virtual ->getPrimaryPlayer()
+            //  move-result-object {vx}
+            val playerRegister = getInstruction<OneRegisterInstruction>(getPlayerIndex + 1).registerA
+
+            // Reuse the params from the original method:
+            //  p0 = ServerInsertedAdBreakState
+            //  p1 = AdBreakTrigger
+            addInstructions(
+                getPlayerIndex + 2,
+                """
```
oSumAtrIX, line 37: Multiline string is not needed here. Simply split the string into new lines for example after the type and the parameters

patches/src/main/kotlin/app/revanced/patches/primevideo/ads/Fingerprints.kt
```diff
@@ -0,0 +1,33 @@
...
+    parameters("Lcom/amazon/avod/fsm/Trigger;")
+    returns("V")
+    opcodes(
+        Opcode.INVOKE_VIRTUAL,
+        Opcode.MOVE_RESULT_OBJECT,
+        Opcode.CONST_4,
+        Opcode.CONST_4
+    )
+    custom { method, classDef ->
+        method.name == "enter" && classDef.type == "Lcom/amazon/avod/media/ads/internal/state/ServerInsertedAdBreakState;"
```
oSumAtrIX, line 18: Is ".type" needed here? ClassDef extends CharSequence

extensions/primevideo/src/main/java/app/revanced/extension/primevideo/ads/SkipAdsPatch.java
```diff
@@ -0,0 +1,25 @@
...
+import com.amazon.avod.media.ads.internal.state.ServerInsertedAdBreakState;
+
+@SuppressWarnings("unused")
+public final class SkipAdsPatch {
+    public static void ServerInsertedAdBreakState_enter(ServerInsertedAdBreakState state, AdBreakTrigger trigger, VideoPlayer player) {
+        AdBreak adBreak = trigger.getBreak();
+
+        if (trigger.getSeekStartPosition() != null)
+            // if scrubbing over ad, seek straight to it
+            player.seekTo(trigger.getSeekTarget().getTotalMilliseconds());
```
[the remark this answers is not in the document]
hoo-dles: When a user manually seeks to a spot in the video that skips over an ad, logic in the app will force the video player timestamp to the start of an ad. After watching the ad, the player will resume to where you were seeking.

In this scenario, when we enter this method, we are at the beginning of the ad and `trigger.getSeekStartPosition()` is set. So force the player position to what the user was originally seeking to.

patches/src/main/kotlin/app/revanced/patches/primevideo/ads/Fingerprints.kt
hoo-dles: Removing `.type` shows a compiler error for the reason @LisoUseInAIKyrios descirbed.

extensions/primevideo/stub/build.gradle.kts
hoo-dles: Build error when removing `compileSdk`:

```
A problem occurred configuring project ':extensions:primevideo:stub'.
> compileSdkVersion is not specified. Please add it to build.gradle
```

patches/src/main/kotlin/app/revanced/patches/primevideo/ads/ExtensionPatch.kt
```diff
@@ -0,0 +1,5 @@
+package app.revanced.patches.primevideo.ads
+
+import app.revanced.patches.shared.misc.extension.sharedExtensionPatch
+
+val sharedExtensionPatch = sharedExtensionPatch("primevideo")
```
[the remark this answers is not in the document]
hoo-dles: Added a hook for the extension. Lmk if the directory structure needs to be modified. There's only a few files, so I just colocated them.

hoo-dles:
@LisoUseInAIKyrios ready if you are

patches/src/main/kotlin/app/revanced/patches/primevideo/ads/ApplicationInitHook.kt
oSumAtrIX, line 1: The file needs to be renamed and moved to its own package like its done for other apps

PR #4829, opened 2025-04-19
feat(Spotify): Add `Sanitize sharing links` patch

### LINKED ISSUE #4826: feat: [Spotify] Add patch: Sanitize sharing links
### Feature description

Add the patch to remove the tracking query parameter from Spotify.

Eg: https://open.spotify.com/track/0gJvyZ5a4d2UQd2ZFWbiGE?si=YCFUrazkSNqfK76YgbpxpA 

The part of '?si=YCF....' should be gone.

### Motivation

The same patch exists for YouTube apps, Reddit and Twitter. Now that Spotify has  again become prominent in ReVanced scene, I would appreciate this patch was added as well.

### Acknowledgements

drobotk:
Closes #4826

### DIFF
--- a/extensions/spotify/src/main/java/app/revanced/extension/spotify/misc/privacy/SanitizeSharingLinksPatch.java
+++ b/extensions/spotify/src/main/java/app/revanced/extension/spotify/misc/privacy/SanitizeSharingLinksPatch.java
@@ -1,0 +1,10 @@
+package app.revanced.extension.spotify.misc.privacy;
+
+@SuppressWarnings("unused")
+public final class SanitizeSharingLinksPatch {
+    private static final String TRACKING_PARAMETER_REGEX = "(?:\\?|&)si=.+?($|&)";
+
+    public static String sanitizeUrl(String url) {
+        return url.replaceFirst(TRACKING_PARAMETER_REGEX, "$1");
+    }
+}
--- a/patches/src/main/kotlin/app/revanced/patches/spotify/misc/privacy/Fingerprints.kt
+++ b/patches/src/main/kotlin/app/revanced/patches/spotify/misc/privacy/Fingerprints.kt
@@ -1,0 +1,7 @@
+package app.revanced.patches.spotify.misc.privacy
+
+import app.revanced.patcher.fingerprint
+
+internal val shareUrlToStringFingerprint = fingerprint {
+    strings("ShareUrl(url=")
+}
--- a/patches/src/main/kotlin/app/revanced/patches/spotify/misc/privacy/SanitizeSharingLinksPatch.kt
+++ b/patches/src/main/kotlin/app/revanced/patches/spotify/misc/privacy/SanitizeSharingLinksPatch.kt
@@ -1,0 +1,35 @@
+package app.revanced.patches.spotify.misc.privacy
+
+import app.revanced.patcher.extensions.InstructionExtensions.addInstructions
+import app.revanced.patcher.patch.bytecodePatch
+import app.revanced.patches.spotify.misc.extension.sharedExtensionPatch
+
+private const val EXTENSION_CLASS_DESCRIPTOR =
+    "Lapp/revanced/extension/spotify/misc/privacy/SanitizeSharingLinksPatch;"
+
+@Suppress("unused")
+val sanitizeSharingLinksPatch = bytecodePatch(
+    name = "Sanitize sharing links",
+    description = "Removes the tracking query parameters from links before they are shared.",
+) {
+    compatibleWith("com.spotify.music")
+
+    dependsOn(sharedExtensionPatch)
+
+    execute {
+        with(shareUrlToStringFingerprint) {
+            classDef.methods.first { it.name == "<init>" }.addInstructions(
+                0,
+                """
+                    invoke-static {p1}, $EXTENSION_CLASS_DESCRIPTOR->sanitizeUrl(Ljava/lang/String;)Ljava/lang/String;
+                    
+                    move-result-object p1
+                    
+                    invoke-static {p4}, $EXTENSION_CLASS_DESCRIPTOR->sanitizeUrl(Ljava/lang/String;)Ljava/lang/String;
+                    
+                    move-result-object p4
+                """
+            )
+        }
+    }
+}
--- a/patches/api/patches.api
+++ b/patches/api/patches.api
@@ -854,6 +854,10 @@ public final class app/revanced/patches/spotify/misc/fix/SpoofSignaturePatchKt {
 	public static final fun getSpoofSignaturePatch ()Lapp/revanced/patcher/patch/BytecodePatch;
 }
 
+public final class app/revanced/patches/spotify/misc/privacy/SanitizeSharingLinksPatchKt {
+	public static final fun getSanitizeSharingLinksPatch ()Lapp/revanced/patcher/patch/BytecodePatch;
+}
+
 public final class app/revanced/patches/spotify/navbar/PremiumNavbarTabPatchKt {
 	public static final fun getPremiumNavbarTabPatch ()Lapp/revanced/patcher/patch/BytecodePatch;
 }


drobotk:
I've now noticed a bug in the regex, need to fix it

drobotk:
Currently does not work on legacy version

patches/src/main/kotlin/app/revanced/patches/spotify/misc/privacy/Fingerprints.kt
```diff
@@ -3,5 +3,14 @@ package app.revanced.patches.spotify.misc.privacy
...
+    strings("ShareUrl", "shareId")
+    custom { method, _ ->
+        method.name == "toString"
+    }
+}
+
+internal val shareUrlConstructorFingerprint = fingerprint {
+    custom { method, _ ->
+        method.name == "<init>"
+    }
```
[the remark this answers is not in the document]
drobotk: Why empty parameter list?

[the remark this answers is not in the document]
drobotk: The constructor has 3 parameters in legacy version, and 4 in latest, how do I handle this?

kitadai31:
Actually, @anddea had already added a similar patch to his RVX fork.
https://github.com/anddea/revanced-patches/commit/dfd7461ad6e0b93ffa5b3cfb8181c12c5a35e8a2
This might be helpful.

Your patch is cleaner in the patches side, but about the extensions code, anddea version is better.
His implementation only uses `String`, so it will never fail.
I think you should borrow his extensions code.

I think the patch description is also anddea version is better.
It clearly explains what the patch does.

(I wonder why he added the Spotify patches to his RVX fork instead of PRing to official ReVanced alongside with the official ReVanced's Unlock Premium patch 🤔)

oSumAtrIX:
The extension code from this PR follows URL specs. This is correct. Arbitrary string manipulation is not recommended

oSumAtrIX:
I wrote this on discord already, but there's a couple of sanitize links patched that may be abstractable into a generic sanitize link patch for code deduplication

Brosssh:
> Actually, @anddea had already added a similar patch to his RVX fork. [anddea/revanced-patches@dfd7461](https://github.com/anddea/revanced-patches/commit/dfd7461ad6e0b93ffa5b3cfb8181c12c5a35e8a2) This might be helpful.
> 
> Your patch is cleaner in the patches side, but about the extensions code, anddea version is better. His implementation only uses `String`, so it will never fail. I think you should borrow his extensions code.
> 
> I think the patch description is also anddea version is better. It clearly explains what the patch does.
> 
> (I wonder why he added the Spotify patches to his RVX fork instead of PRing to official ReVanced alongside with the official ReVanced's Unlock Premium patch 🤔)

The extension method done by anddea is basically wrong tho.. it's simply returning everything before `?si=`. The problem is Spotify sometimes actually add additional context to a shareable link (for instance in which playlist is the song that's being shared). That should not be removed from the shareable URL (in my opinion at least).

kitadai31:
> The problem is Spotify sometimes actually add additional context to a shareable link

I didn't know, sorry
That was a serious issue... Thank you

anddea:
> it's simply returning everything before ?si=

That's true, I hadn't considered that aspect. But I've never come across links like that. Do you have any examples?

> patch is cleaner

I can simplify my patch. I overcomplicated the fingerprint and invokeDirectIndex during testing to match more cases and ensure I hit the right spot on the first try. I just never simplified it afterward because there wasn’t really a reason to, but yeah, it can be drastically simplified. I even have both the fingerprint and the patch in a single file, which might be considered as not the best practice by some people.

But why are there two instructions, though? I’ve been using just one, and I haven’t run into any issues so far. Am I missing something?

> (I wonder why he added the Spotify patches to his RVX fork instead of PRing to official ReVanced alongside with the official ReVanced's Unlock Premium patch 🤔)

I don't mind if anyone moves all my patches here or even rewrites it in a better way. I just don't want to make the same thing twice, I experiment in my repo where I can push immediately.

drobotk:
Can this be merged for now, or should I convert this to a generic patch still in this same PR?

Brosssh:
> > it's simply returning everything before ?si=
> 
> That's true, I hadn't considered that aspect. But I've never come across links like that. Do you have any examples?

`https://open.spotify.com/track/6H6UoUKz8OVaLXqQbZgbPK?si=WHmwIx3tTVOrwfeWGFn6EQ&context=spotify%3Aalbum%3A3bwpDqNK4LGNW9ZkfltZ6I`
This for example. I just pressed share for a song inside its album.

anddea:
> &context=spotify%3Aalbum%3A3bwpDqNK4LGNW9ZkfltZ6I

Not sure if this part is really necessary, spamming the URL just makes it even longer (since a Spotify link already opens the album directly). But I'll merge these changes into my repo anyway.

Genuine question though, why inject two separate calls to sanitizeUrl? Even though the end user won't notice any big impact, creating the Uri, Builder, and other objects is already less performant than raw string manipulation, and here it's doing all that twice. Seems like overhead that should be avoided if possible.

Nuckyz:
> > &context=spotify%3Aalbum%3A3bwpDqNK4LGNW9ZkfltZ6I
> 
> Not sure if this part is really necessary, spamming the URL just makes it even longer (since a Spotify link already opens the album directly). But I'll merge these changes into my repo anyway.
> 
> Genuine question though, why inject two separate calls to sanitizeUrl? Even though the end user won't notice any big impact, creating the Uri, Builder, and other objects is already less performant than raw string manipulation, and here it's doing all that twice. Seems like overhead that should be avoided if possible.

That seems like a huge micro optimization which shouldn't be considered. The user will click little times in the share button, that won't have any issues.

anddea:
> huge micro optimization
> The user will click little times in the share button, that won't have any issues.

Sure (why huge though?)

> the end user won't notice any big impact

Still, with one injection in my repo, there were no issues (if I'm not missing anything, never got any reports). So if possible, then why?

drobotk:
I wasn't sure whether fullUrl should be sanitized too, so I figured its better to be safe than sorry

anddea:
> I think it's more long term stable to instead modify the url passed to the share intent, rather than broadcast modified urls to Spotify that are missing data.

This is quite complicated, there are many targets involved: Copy Link, SMS, WhatsApp, Instagram, X, and "More" (which opens the Android share sheet for additional apps).

In the end, does it really matter if Spotify receives these modified URLs? If they notice and decide to change the sharing logic, they could just as easily check whatever method is being used here and update their code in a future release.

> if quick share targets are added they won't be modified until the patch is updated

With the current implementation (at least in my patch), URLs are already being sanitized for all share targets, and this comes at a relatively low cost. Even if Spotify breaks this patch by altering the output format or whatever, the impact is less severe compared to the effort required to implement a more complex solution that could still be broken just as easily.

patches/src/main/kotlin/app/revanced/patches/spotify/misc/privacy/Fingerprints.kt
```diff
@@ -0,0 +1,43 @@
...
+    accessFlags(AccessFlags.PUBLIC, AccessFlags.FINAL)
+    returns("Ljava/lang/Object;")
+    parameters("Ljava/lang/Object;")
+    strings("clipboard", "Spotify Link")
+    custom { method, _ ->
+        method.name == "invokeSuspend"
+    }
+}
+
+internal val shareCopyUrlLegacyFingerprint = fingerprint {
```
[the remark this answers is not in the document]
Nuckyz: Just checked, it's not easy

patches/src/main/kotlin/app/revanced/patches/spotify/misc/privacy/Fingerprints.kt
```diff
@@ -0,0 +1,43 @@
...
+internal val androidShareSheetUrlFormatterFingerprint = fingerprint {
+    accessFlags(AccessFlags.PUBLIC, AccessFlags.STATIC)
+    returns("Ljava/lang/String;")
+    parameters("L", "Ljava/lang/String;")
+    literal {
+        '\n'.code.toLong()
+    }
+}
+
+internal val androidShareSheetUrlFormatterLegacyFingerprint = fingerprint {
```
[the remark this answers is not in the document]
Nuckyz: Also not easy, ignore it

extensions/spotify/src/main/java/app/revanced/extension/spotify/misc/privacy/SanitizeSharingLinksPatch.java
```diff
@@ -0,0 +1,43 @@
...
+@SuppressWarnings("unused")
+public final class SanitizeSharingLinksPatch {
+
+    /**
+     * Parameters that are considered undesirable and should be stripped away.
+     */
+    private static final List<String> SHARE_PARAMETERS_TO_REMOVE = List.of(
+            "si", // Share tracking parameter.
+            "utm_source" // Share source, such as "copy-link".
+    );
```
oSumAtrIX, line 18: This should be a set by semantic.

patches/src/main/kotlin/app/revanced/patches/spotify/misc/extension/ExtensionPatch.kt
```diff
@@ -17,5 +18,11 @@ val sharedExtensionPatch = bytecodePatch {
     execute {
         IS_SPOTIFY_LEGACY_APP_TARGET = mainActivityOnCreateHook.fingerprint
             .originalClassDef.type == SPOTIFY_MAIN_ACTIVITY_LEGACY
+
+        if (IS_SPOTIFY_LEGACY_APP_TARGET) {
+            Logger.getLogger(this::class.java.name).warning(
+                "Patching a legacy Spotify version. Patch functionality may be limited."
```
oSumAtrIX, line 24: How is the extension patch limited when applied to old versions?

patches/src/main/kotlin/app/revanced/patches/spotify/misc/privacy/Fingerprints.kt
```diff
@@ -0,0 +1,43 @@
...
+
+internal val shareCopyUrlFingerprint = fingerprint {
+    accessFlags(AccessFlags.PUBLIC, AccessFlags.FINAL)
+    returns("Ljava/lang/Object;")
+    parameters("Ljava/lang/Object;")
+    strings("clipboard", "Spotify Link")
+    custom { method, _ ->
+        method.name == "invokeSuspend"
+    }
+}
```
oSumAtrIX, line 15: Are all parameters necessary besides the string and custom?

patches/src/main/kotlin/app/revanced/patches/spotify/misc/privacy/Fingerprints.kt
```diff
@@ -0,0 +1,43 @@
...
+    accessFlags(AccessFlags.PUBLIC, AccessFlags.FINAL)
+    returns("Ljava/lang/Object;")
+    parameters("Ljava/lang/Object;")
+    strings("clipboard", "createNewSession failed")
+    custom { method, _ ->
+        method.name == "apply"
+    }
+}
+
+internal val androidShareSheetUrlFormatterFingerprint = fingerprint {
```
oSumAtrIX, line 27: This should be named formatAndroidShareSheetUrlFingerprint and likewise the other fingerprints.

patches/src/main/kotlin/app/revanced/patches/spotify/misc/privacy/SanitizeSharingLinksPatch.kt
```diff
@@ -0,0 +1,71 @@
...
+        val copyFingerprint = if (IS_SPOTIFY_LEGACY_APP_TARGET) {
+            shareCopyUrlLegacyFingerprint
+        } else {
+            shareCopyUrlFingerprint
+        }
+
+        copyFingerprint.method.apply {
+            val index = indexOfFirstInstructionOrThrow {
+                val reference = getReference<MethodReference>()
+                reference?.name == "newPlainText"
```
oSumAtrIX, line 39:
```suggestion
                getReference<MethodReference>()?.name == "newPlainText"

```

patches/src/main/kotlin/app/revanced/patches/spotify/misc/privacy/SanitizeSharingLinksPatch.kt
```diff
@@ -0,0 +1,71 @@
...
+                "sanitizeUrl(Ljava/lang/String;)Ljava/lang/String;"
+
+        val copyFingerprint = if (IS_SPOTIFY_LEGACY_APP_TARGET) {
+            shareCopyUrlLegacyFingerprint
+        } else {
+            shareCopyUrlFingerprint
+        }
+
+        copyFingerprint.method.apply {
+            val index = indexOfFirstInstructionOrThrow {
```
oSumAtrIX, line 37: Call this newPlainTextInvokeIndex

patches/src/main/kotlin/app/revanced/patches/spotify/misc/extension/ExtensionPatch.kt
oSumAtrIX: This is matter of the other patches, not the extension patch. The correct location for this notice is the premium unlock patch.

extensions/spotify/src/main/java/app/revanced/extension/spotify/misc/privacy/SanitizeSharingLinksPatch.java
```diff
@@ -0,0 +1,43 @@
...
+            }
+
+            return builder.build().toString();
+        } catch (Exception ex) {
+            Logger.printException(() -> "sanitizeUrl failure", ex);
+
+            return url;
+        }
+    }
+}
```
Nuckyz, line 43:
```suggestion
}

```

patches/src/main/kotlin/app/revanced/patches/spotify/misc/privacy/SanitizeSharingLinksPatch.kt
```diff
@@ -0,0 +1,70 @@
...
+
+        shareSheetFingerprint.method.addInstructions(
+            0,
+            """
+                invoke-static { $shareUrlParameter }, $extensionMethodDescriptor
+                move-result-object $shareUrlParameter
+            """
+        )
+    }
+}
```
Nuckyz, line 70:
```suggestion
}

```

Nuckyz (changes requested):
Nitpick to match all the other patches in Spotify, which all include new lines at the end of the files

PR #4850, opened 2025-04-24
feat(scrl): Add unlock premium feature

Hug0Vincent:
Premium unlock feature for the [SCRL app](https://scrl.com). 

It's my first contribution, let me know if I need to change something :)


Hug0Vincent:
Hello @LisoUseInAIKyrios sorry if it's not ok with your policy but what's the difference with the patch for the com.backdrops.wallpapers or the com.zombodroid.MemeGenerator ? Just asking to understand what's acceptable and what's not :)

oSumAtrIX:
Basically those were added in the past, before the guidelines came into effect. See the guidelines.md file. However, you can publish your patches and the new version of manager will be able to handle multiple sources of patches at once which means people can add yours as well, decentralized outside of our domain or policies.

Hug0Vincent:
Alright, perfect thank you for the information

PR #4920, opened 2025-05-07
fix(TikTok): Update patches to latest version

Nuckyz:
TODO: Find a way to inject the settings button

PR also does a lot of clean-up

### DIFF
--- a/extensions/tiktok/src/main/java/app/revanced/extension/tiktok/Utils.java
+++ b/extensions/tiktok/src/main/java/app/revanced/extension/tiktok/Utils.java
@@ -12,7 +12,6 @@ import androidx.annotation.ColorInt;
 import app.revanced.extension.shared.settings.StringSetting;
 
 public class Utils {
-
     private static final long[] DEFAULT_MIN_MAX_VALUES = {0L, Long.MAX_VALUE};
 
     // Edit: This could be handled using a custom Setting<Long[]> class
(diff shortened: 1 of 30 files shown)


Nuckyz:
I need assist on this one. The new code for UI which creates the settings is completely automatically generated by their tools, and each component is composed by very specific code generated exactly for it. It seems extremely hard to use their own components now to add the settings button, so a different approach would probably be better. However, I am sure not how a different approach could be done, as I dont have a lot of experience with UI in specific.

Nuckyz:
One approach could be perhaps modifying one of the existing buttons

Nuckyz:
I have some new ideas but haven't tried yet. I will come back here once I try.

Nuckyz:
Been a little bit busy in the past days, I will continue this as soon as I'm able to.

pikaczu787:
@Nuckyz are you gay?

Nuckyz:
> @Nuckyz Any update?

Busy with Spotify for now, this will have to wait.

Nuckyz:
> I meant, why not repose that button so it opens the ReVanced settings. Can pick a button that isn't important.
> 
> That stock button would lose its original functionality, but maybe it could be restored by calling into the original code/listener (add a replacement button in the ReVanced dialog that calls the original code).

Alright, I'll keep in that mind for the next time I resume this

Nuckyz:
> can confirm this branch works fine on TikTik 39.8.4 and rebased on top of the latest release tag v5.27.0, the lack of settings doesn't really bother me, just changed them in the source thanks!

Some patches still crash so be aware

AykhanUV:
@LisoUseInAIKyrios without settings button you can't configure most of the patches, and that makes them unusable too

oSumAtrIX:
image

PR #4604, opened 2025-03-17
feat(Angulus): Add `Hide ads` patch

dtricks:
Hides ads in the [Angulus](https://play.google.com/store/apps/details?id=com.drinkplusplus.angulus) app

PR #4927, opened 2025-05-08
feat(Angulus): Adding `Disable license check` patch

### LINKED ISSUE #4776: bug(Angulus): `Hide ads` succeeded but the app won't open
### Bug description

It was redirected to the Play Store.
`v5.0.16` and `v5.0.20` same result.

Image

### Error logs

```shell
- Device Info
ReVanced Manager: 1.24.1-dev.3
Model: Redmi Note 9 Pro
Android version: 12
Supported architectures: arm64-v8a, armeabi-v7a, armeabi
Root permissions: No

- Patch Info
App: com.drinkplusplus.angulus v5.0.20 (Suggested: Any)
Patches version: v5.19.1
Patches added: Default
Patches removed: None
Default patch options changed: None

- Settings
Allow changing patch selection: false
Version compatibility check: true
Show universal patches: false
Patches source: revanced/revanced-patches

- Logs
Reading APK
Decoding app manifest
Loading patches
Initializing lookup maps
Executing patches
Applied 1 patches
Aligning APK
Signing APK
Patched APK
```

### Solution

_No response_

### Additional context

_No response_

### Acknowledgements

hoo-dles:
Just a simple, little patch to bypass the client-side validation of the Google Play Licensing response. This should work for any app that only does the client validation (if we want to move the code to a shared location.)

Fixes #4776

### DIFF
--- a/patches/src/main/kotlin/app/revanced/patches/angulus/misc/license/DisableLicenseCheckPatch.kt
+++ b/patches/src/main/kotlin/app/revanced/patches/angulus/misc/license/DisableLicenseCheckPatch.kt
@@ -1,0 +1,21 @@
+package app.revanced.patches.angulus.misc.license
+
+import app.revanced.patcher.extensions.InstructionExtensions.addInstruction
+import app.revanced.patcher.patch.bytecodePatch
+import app.revanced.util.returnEarly
+
+@Suppress("unused")
+val disableLicenseCheckPatch = bytecodePatch(
+    name = "Disable license check",
+    description = "Disable client-side license check to prevent app from closing on startup."
+) {
+    compatibleWith("com.drinkplusplus.angulus"("5.0.20"))
+
+    execute {
+        // Set first parameter (responseCode) to 0.
+        processLicenseResponseFingerprint.method.addInstruction(0, "const/4 p1, 0x0")
+
+        // Short-circuit the license response validation.
+        validateLicenseResponseFingerprint.method.returnEarly();
+    }
+}
--- a/patches/src/main/kotlin/app/revanced/patches/angulus/misc/license/Fingerprints.kt
+++ b/patches/src/main/kotlin/app/revanced/patches/angulus/misc/license/Fingerprints.kt
@@ -1,0 +1,17 @@
+package app.revanced.patches.angulus.misc.license
+
+import app.revanced.patcher.fingerprint
+
+internal val processLicenseResponseFingerprint = fingerprint {
+    custom { method, classDef ->
+        classDef.type == "Lcom/pairip/licensecheck/LicenseClient;" &&
+                method.name == "processResponse"
+    }
+}
+
+internal val validateLicenseResponseFingerprint = fingerprint {
+    custom { method, classDef ->
+        classDef.type == "Lcom/pairip/licensecheck/ResponseValidator;" &&
+                method.name == "validateResponse"
+    }
+}
--- a/patches/api/patches.api
+++ b/patches/api/patches.api
@@ -144,6 +144,10 @@ public final class app/revanced/patches/angulus/ads/RemoveAdsPatchKt {
 	public static final fun getAngulusPatch ()Lapp/revanced/patcher/patch/BytecodePatch;
 }
 
+public final class app/revanced/patches/angulus/misc/license/DisableLicenseCheckPatchKt {
+	public static final fun getDisableLicenseCheckPatch ()Lapp/revanced/patcher/patch/BytecodePatch;
+}
+
 public final class app/revanced/patches/backdrops/misc/pro/ProUnlockPatchKt {
 	public static final fun getProUnlockPatch ()Lapp/revanced/patcher/patch/BytecodePatch;
 }


patches/src/main/kotlin/app/revanced/patches/angulus/misc/license/DisableLicenseCheckPatch.kt
```diff
@@ -0,0 +1,21 @@
...
+@Suppress("unused")
+val disableLicenseCheckPatch = bytecodePatch(
+    name = "Disable license check",
+    description = "Disable client-side license check to prevent app from closing on startup."
+) {
+    compatibleWith("com.drinkplusplus.angulus"("5.0.20"))
+
+    execute {
+        // Set first parameter (responseCode) to 0.
+        processLicenseResponseFingerprint.method.addInstruction(0, "const/4 p1, 0x0")
```
Nuckyz, line 16: Should explain what 0 means as the response code and how it affects the code

patches/src/main/kotlin/app/revanced/patches/angulus/misc/license/Fingerprints.kt
```diff
@@ -0,0 +1,17 @@
...
+                method.name == "processResponse"
+    }
+}
+
+internal val validateLicenseResponseFingerprint = fingerprint {
+    custom { method, classDef ->
+        classDef.type == "Lcom/pairip/licensecheck/ResponseValidator;" &&
+                method.name == "validateResponse"
+    }
+}
```
Nuckyz, line 17:
```suggestion
}

```

Nuckyz (changes requested):

patches/src/main/kotlin/app/revanced/patches/angulus/misc/license/DisableLicenseCheckPatch.kt
```diff
@@ -0,0 +1,21 @@
...
+import app.revanced.patcher.extensions.InstructionExtensions.addInstruction
+import app.revanced.patcher.patch.bytecodePatch
+import app.revanced.util.returnEarly
+
+@Suppress("unused")
+val disableLicenseCheckPatch = bytecodePatch(
+    name = "Disable license check",
+    description = "Disable client-side license check to prevent app from closing on startup."
+) {
+    compatibleWith("com.drinkplusplus.angulus"("5.0.20"))
```
Nuckyz, line 12:
```suggestion
    compatibleWith("com.drinkplusplus.angulus")
```

Considering this patch is more specific to pairip than Angulus, it's unlikely to change

Nuckyz (changes requested):
I would also take a look at changing the daily limit patch to returnEarly(0) as the method returns an int, and calling returnEarly without a value seems kinda weird

dtricks:
> I would also take a look at changing the daily limit patch to returnEarly(0) as the method returns an int, and calling returnEarly without a value seems kinda weird

https://github.com/ReVanced/revanced-patches/pull/4604#discussion_r1999509526

I find `returnEarly(0)` more readable as well

hoo-dles:
> I would also take a look at changing the daily limit patch to returnEarly(0) as the method returns an int, and calling returnEarly without a value seems kinda weird

@oSumAtrIX We currently don't have a helper method for `returnEarly()` that accepts primitive types other than boolean. `returnEarly(false)` does not seem any more readable in this instance since the original method returns an integer. Do we want to extend the existing helpers, or just leave the code as is?

hoo-dles:
> > The constant size is hardcoded to a 4-bit const/4, how do you want to handle that?
> 
> Why not replace the instruction with regular `const`? If the 4-bit limit is the issue you're speaking of.

That's definitely a possibility. Although I don't think that would work for a string literal (where you'd use `const-string`), which is confusing if the function signature accepts a `string`. I figured I would ask the team considering this a widely used helper throughout the patches project.

hoo-dles:
> Currently `returnEarly` (and it's similar `returnLate`) don't work with strings but feel free to add it to the existing util methods.
> 
> Instead of just boolean and string overloads, the utils method could be overloaded with all primitive types plus String, as that gives better type safety since the parameter type would need to match the method being patched. Then for the implementation just modify the existing private util method and have all the overloadeds call onto it.

Pushed new commits that move pairip patch to shared location, and added `returnEarly()` and `returnLate()` overloads for primitives.

 I didn't do an overload for `String` because all the Smali code I've seen treats strings as the Java object. Wasn't sure if `const-string` accepted a variable size string literal (opcode list says `string_id` param) and played nicely with a return type of `Ljava/lang/String;`.

patches/src/main/kotlin/app/revanced/patches/shared/misc/pairip/license/DisableLicenseCheckPatch.kt
```diff
@@ -0,0 +1,20 @@
+package app.revanced.patches.shared.misc.pairip.license
+
+import app.revanced.patcher.extensions.InstructionExtensions.addInstruction
+import app.revanced.patcher.patch.bytecodePatch
+import app.revanced.util.returnEarly
+
+@Suppress("unused")
+val disableLicenseCheckPatch = bytecodePatch(
+    name = "Disable license check",
+    description = "Disable Play Integrity Protect (pairip) client-side license check."
```
Brosssh, line 10: Description should be improved in my opinion. Basically every pairip version is based on `libpairipcore.so`, and that cannot be disabled (not by this patch, at least). May be worth noting it in the patch description, this patch is only bypassing the smali license check, but if the app has native checks, it won't open anyway.

Brosssh (changes requested):

patches/src/main/kotlin/app/revanced/util/BytecodeUtils.kt
```diff
@@ -730,26 +738,123 @@ fun BytecodePatchContext.forEachLiteralValueInstruction(
...
+ * Overrides all return statements with a constant `Double` value.
+ * All method code is executed the same as unpatched.
+ *
+ * @see returnEarly
+ */
+internal fun MutableMethod.returnLate(value: Double) = overrideReturnValue(value.toHexString(), true)
+
+private fun MutableMethod.overrideReturnValue(value: String, returnLate: Boolean) {
     val instructions = when (returnType.first()) {
+        // If return type is an object, always return null.
```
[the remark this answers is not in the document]
hoo-dles: Yeah, I just kept it for backwards compatibility if it was used anywhere.

patches/src/main/kotlin/app/revanced/patches/shared/misc/pairip/license/DisableLicenseCheckPatch.kt
hoo-dles: I'll change it to whatever language someone suggests :P

patches/src/main/kotlin/app/revanced/util/BytecodeUtils.kt
```diff
@@ -759,9 +864,30 @@ private fun MutableMethod.overrideReturnValue(bool: Boolean, returnLate: Boolean
...
+                const v0, $value
+                return v0
+            """
+        }
+
+        'J', 'D' -> {
             """
-                const/4 v0, $const
+                const-wide v0, $value
                 return v0
```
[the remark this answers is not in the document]
hoo-dles: good catch

patches/src/main/kotlin/app/revanced/util/BytecodeUtils.kt
```diff
@@ -730,26 +738,123 @@ fun BytecodePatchContext.forEachLiteralValueInstruction(
...
+ */
+internal fun MutableMethod.returnLate(value: Int) = overrideReturnValue(value.toHexString(), true)
+
+/**
+ * Overrides all return statements with a constant `Long` value.
+ * All method code is executed the same as unpatched.
+ *
+ * @see returnEarly
+ */
+internal fun MutableMethod.returnLate(value: Long) = overrideReturnValue(value.toHexString(), true)
```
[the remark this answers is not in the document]
hoo-dles: > Only the boolean type needs to be converted to hex. Double/floats can be used as-is and the assembler converts if needed.
> 
> Returnearly/late can be public and not internal.

Same with Short/Int/Long (and negatives)? I assume Char at least needs to be converted to a Short.

patches/src/main/kotlin/app/revanced/util/BytecodeUtils.kt
```diff
@@ -729,43 +737,215 @@ fun BytecodePatchContext.forEachLiteralValueInstruction(
...
 
+private const val RETURN_TYPE_MISMATCH = "Mismatch between override type and Method return type"
+
+/**
+ * Overrides the first instruction of a method with a constant `Boolean` return value.
+ * None of the method code will ever execute.
+ */
+fun MutableMethod.returnEarly(value: Boolean = false) {
+    val returnType = returnType.first();
+    check(returnType == 'Z' || (!value && (returnType == 'V' || returnType == 'L'))) { RETURN_TYPE_MISMATCH }
```
hoo-dles, line 748: Per @LisoUseInAIKyrios :

> Could also check array return type:
> `returnType == 'V' || returnType == 'L' || return type == '['`

Issue #4906, opened 2025-05-04
feat(Spotify): Allow using custom api as a lyrics provider


YT-Advanced:
### Feature description

- Just replace the spotify api with our [custom api](https://github.com/Natoune/SpotifyMobileLyricsAPI) using the idea from this [https://github.com/Natoune/SpotifyMobileLyricsAPI/blob/main/scripts/xmanager.patch.js]

### Motivation

MusixMatch sometimes doesn't have lyrics for our wanted songs

### Acknowledgements

- [x] I have checked all open and closed feature requests and this is not a duplicate
- [x] I have chosen an appropriate title.
- [x] All requested information has been provided properly.

bybrl:
I believe that using a method to lyrics like innertune would be a good way to go about this. Eg. Choosing where to fetch lyrics from

Rairof:
https://github.com/ReVanced/revanced-patches/issues/4906#issuecomment-2849370055
It would be better to have an automated selection of lyrics where if a song lyric isn't available in Spotify, it switches automatically to a different provider till it finds lyrics of the song.

oSumAtrIX:
The respective backend endpoint, set via patch options can do that. It's easier than writing a patch

PR #4937, opened 2025-05-10
feat(Spotify): Custom lyrics patch

Brosssh:
The user will enter an URL. This URL will be used when fetching lyrics.
The implemented logic is the same as https://github.com/Natoune/SpotifyMobileLyricsAPI/blob/main/scripts/xmanager.patch.js.

Closes https://github.com/ReVanced/revanced-patches/issues/4906

### DIFF
--- a/patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
+++ b/patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
@@ -1,0 +1,133 @@
+package app.revanced.patches.spotify.misc.lyrics
+
+import app.revanced.patcher.Fingerprint
+import app.revanced.patcher.extensions.InstructionExtensions.addInstruction
+import app.revanced.patcher.extensions.InstructionExtensions.addInstructions
+import app.revanced.patcher.extensions.InstructionExtensions.getInstruction
+import app.revanced.patcher.extensions.InstructionExtensions.instructions
+import app.revanced.patcher.extensions.InstructionExtensions.replaceInstruction
+import app.revanced.patcher.patch.bytecodePatch
+import app.revanced.patcher.patch.stringOption
+import app.revanced.patcher.util.proxy.mutableTypes.MutableMethod.Companion.toMutable
+import app.revanced.util.indexOfFirstInstructionOrThrow
+import app.revanced.util.indexOfFirstInstructionReversedOrThrow
+import com.android.tools.smali.dexlib2.Opcode
+import com.android.tools.smali.dexlib2.builder.MutableMethodImplementation
+import com.android.tools.smali.dexlib2.builder.instruction.BuilderInstruction21t
+import com.android.tools.smali.dexlib2.builder.instruction.BuilderInstruction35c
+import com.android.tools.smali.dexlib2.iface.instruction.FiveRegisterInstruction
+import com.android.tools.smali.dexlib2.iface.instruction.OneRegisterInstruction
+import com.android.tools.smali.dexlib2.immutable.ImmutableMethod
+import com.android.tools.smali.dexlib2.immutable.reference.ImmutableMethodReference
+
+
+@Suppress("unused")
+val customLyricsPatch = bytecodePatch(
+    name = "Custom lyrics provider",
+    description = "Make Spotify use a custom lyrics provider.",
+) {
+    compatibleWith("com.spotify.music")
+
+    val lyricsProviderUrlOption = stringOption(
+        key = "lyricsProviderUrl",
+        title = "Lyrics provider URL",
+        description = "The custom lyrics provider URL.",
+        required = true,
+    )
+
+    execute {
+        val lyricsProviderUrl by lyricsProviderUrlOption()
+        var patchedClientMethod: ImmutableMethod?
+
+        fun addLyricsProviderToValidHosts(fingerprint: Fingerprint) {
+            with(fingerprint) {
+                val validUrlAssignmentIndex = method.indexOfFirstInstructionOrThrow(
+                    stringMatches!!.first().index, Opcode.CONST_STRING
+                )
+
+                val urlComparisonIndex = method.indexOfFirstInstructionOrThrow(
+                    validUrlAssignmentIndex,
+                    Opcode.INVOKE_VIRTUAL
+                )
+
+                val ifValidUrlIndex = method.indexOfFirstInstructionOrThrow(
+                    urlComparisonIndex,
+                    Opcode.IF_NEZ
+                )
+
+                val label = method.getInstruction<BuilderInstruction21t>(ifValidUrlIndex).target
+
+                val (stringRegister, compareToRegister) = method.getInstruction<FiveRegisterInstruction>(urlComparisonIndex).let { it.registerC to it.registerD }
+
+                method.addInstructions(ifValidUrlIndex + 1,
+                    """
+                    const-string v$stringRegister, "$lyricsProviderUrl"
+                    
+                    invoke-virtual {v$compareToRegister, v$stringRegister}, Ljava/lang/Object;->equals(Ljava/lang/Object;)Z
+                    
+                    move-result v$stringRegister
+                """
+                )
+
+                method.addInstruction(ifValidUrlIndex + 1 + 3,
+                    BuilderInstruction21t(Opcode.IF_NEZ, stringRegister, label))
+            }
+        }
+
+        addLyricsProviderToValidHosts(oauthHostnameCheckFingerprint)
+        addLyricsProviderToValidHosts(webgateHostnameCheckFingerprint)
+
+        val clientBuilderMethodName = "getCustomLyricsProviderClient"
+
+        with(clientBuilderFingerprint) {
+            val clientStringAssignmentIndex = method.indexOfFirstInstructionOrThrow(
+                stringMatches!!.first().index,
+                Opcode.CONST_STRING,
+            )
+
+            val urlRegister = method.getInstruction<OneRegisterInstruction>(clientStringAssignmentIndex).registerA
+
+            patchedClientMethod = ImmutableMethod(
+                method.definingClass,
+                clientBuilderMethodName,
+                method.parameters,
+                method.returnType,
+                method.accessFlags,
+                null,
+                null,
+                MutableMethodImplementation(6),
+            )
+
+            classDef.methods.add(
+                patchedClientMethod!!.toMutable().apply {
+                    addInstructions(method.instructions)
+
+                    replaceInstruction(clientStringAssignmentIndex, "const-string v$urlRegister, \"$lyricsProviderUrl\"")
+                }
+            )
+        }
+
+        with(executeFingerprint.method) {
+            val getLyricsClientIndex = indexOfFirstInstructionReversedOrThrow(Opcode.INVOKE_STATIC)
+            val getLyricsClientInstruction = getInstruction<BuilderInstruction35c>(getLyricsClientIndex)
+
+            replaceInstruction(getLyricsClientIndex,
+                BuilderInstruction35c(
+                    Opcode.INVOKE_STATIC,
+                    getLyricsClientInstruction.registerCount,
+                    getLyricsClientInstruction.registerC,
+                    getLyricsClientInstruction.registerD,
+                    getLyricsClientInstruction.registerE,
+                    getLyricsClientInstruction.registerF,
+                    getLyricsClientInstruction.registerG,
+                    ImmutableMethodReference(
+                        patchedClientMethod!!.definingClass,
+                        clientBuilderMethodName,
+                        patchedClientMethod!!.parameters,
+                        patchedClientMethod!!.returnType
+                    )
+                )
+            )
+        }
+    }
+}
--- a/patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/Fingerprints.kt
+++ b/patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/Fingerprints.kt
@@ -1,0 +1,29 @@
+package app.revanced.patches.spotify.misc.lyrics
+
+import app.revanced.patcher.Fingerprint
+import app.revanced.patcher.fingerprint
+import app.revanced.patcher.patch.BytecodePatchContext
+import com.android.tools.smali.dexlib2.Opcode
+
+internal val clientBuilderFingerprint = fingerprint {
+    strings("spclient.wg.spotify.com")
+    parameters("Lokhttp3/OkHttpClient;", "Lcom/fasterxml/jackson/databind/ObjectMapper;", "L", "Lio/reactivex/rxjava3/core/Scheduler;")
+}
+
+internal val oauthHostnameCheckFingerprint = fingerprint {
+    strings("spclient.wg.spotify.com")
+    parameters("Lokhttp3/Request;")
+    returns("Z")
+}
+
+internal val webgateHostnameCheckFingerprint = fingerprint {
+    strings("spclient.wg.spotify.com", "Authorization")
+}
+
+context(BytecodePatchContext)
+internal val executeFingerprint: Fingerprint
+    get() = fingerprint {
+        returns(clientBuilderFingerprint.originalMethod.returnType)
+        parameters()
+        opcodes(Opcode.CHECK_CAST)
+    }
--- a/patches/api/patches.api
+++ b/patches/api/patches.api
@@ -868,6 +868,10 @@ public final class app/revanced/patches/spotify/misc/fix/SpoofSignaturePatchKt {
 	public static final fun getSpoofSignaturePatch ()Lapp/revanced/patcher/patch/BytecodePatch;
 }
 
+public final class app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatchKt {
+	public static final fun getCustomLyricsPatch ()Lapp/revanced/patcher/patch/BytecodePatch;
+}
+
 public final class app/revanced/patches/spotify/misc/privacy/SanitizeSharingLinksPatchKt {
 	public static final fun getSanitizeSharingLinksPatch ()Lapp/revanced/patcher/patch/BytecodePatch;
 }


patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
```diff
@@ -0,0 +1,133 @@
...
+
+
+@Suppress("unused")
+val customLyricsPatch = bytecodePatch(
+    name = "Custom lyrics provider",
+    description = "Make Spotify use a custom lyrics provider.",
+) {
+    compatibleWith("com.spotify.music")
+
+    val lyricsProviderUrlOption = stringOption(
```
Brosssh, line 31: Never used this before and I'm not sure to test the UI. Please give it a look.

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
```diff
@@ -0,0 +1,133 @@
...
+        key = "lyricsProviderUrl",
+        title = "Lyrics provider URL",
+        description = "The custom lyrics provider URL.",
+        required = true,
+    )
+
+    execute {
+        val lyricsProviderUrl by lyricsProviderUrlOption()
```
Brosssh, line 39: How can we do input check? Expected is an url without `https://` and without `/`. Should we clean the URL or just show the user the expected URL format?

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
```diff
@@ -0,0 +1,133 @@
...
+                    replaceInstruction(clientStringAssignmentIndex, "const-string v$urlRegister, \"$lyricsProviderUrl\"")
+                }
+            )
+        }
+
+        with(executeFingerprint.method) {
+            val getLyricsClientIndex = indexOfFirstInstructionReversedOrThrow(Opcode.INVOKE_STATIC)
+            val getLyricsClientInstruction = getInstruction<BuilderInstruction35c>(getLyricsClientIndex)
+
+            replaceInstruction(getLyricsClientIndex,
```
Brosssh, line 114: Pretty sure there is a simplier way for this but couldn't figure it out

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/Fingerprints.kt
```diff
@@ -0,0 +1,29 @@
...
+    parameters("Lokhttp3/Request;")
+    returns("Z")
+}
+
+internal val webgateHostnameCheckFingerprint = fingerprint {
+    strings("spclient.wg.spotify.com", "Authorization")
+}
+
+context(BytecodePatchContext)
+internal val executeFingerprint: Fingerprint
```
Brosssh, line 24: @oSumAtrIX said that a `backing field to cache` should be added. I'm not sure what this is so I let it for you :) .

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
```diff
@@ -0,0 +1,133 @@
...
+import com.android.tools.smali.dexlib2.iface.instruction.FiveRegisterInstruction
+import com.android.tools.smali.dexlib2.iface.instruction.OneRegisterInstruction
+import com.android.tools.smali.dexlib2.immutable.ImmutableMethod
+import com.android.tools.smali.dexlib2.immutable.reference.ImmutableMethodReference
+
+
+@Suppress("unused")
+val customLyricsPatch = bytecodePatch(
+    name = "Custom lyrics provider",
+    description = "Make Spotify use a custom lyrics provider.",
```
oSumAtrIX, line 27:
```suggestion
    description = "Changes the lyrics provider to a custom one.",
```

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
```diff
@@ -0,0 +1,133 @@
...
+import com.android.tools.smali.dexlib2.builder.instruction.BuilderInstruction35c
+import com.android.tools.smali.dexlib2.iface.instruction.FiveRegisterInstruction
+import com.android.tools.smali.dexlib2.iface.instruction.OneRegisterInstruction
+import com.android.tools.smali.dexlib2.immutable.ImmutableMethod
+import com.android.tools.smali.dexlib2.immutable.reference.ImmutableMethodReference
+
+
+@Suppress("unused")
+val customLyricsPatch = bytecodePatch(
+    name = "Custom lyrics provider",
```
oSumAtrIX, line 26: The patch should be named according to the field or vice versa. Ideally "Change lyrics provider"

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
oSumAtrIX: In the option above you can pass a lambda to check the option on value input:

stringOption(...) { validate here }

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/Fingerprints.kt
oSumAtrIX, line 24: Just inline this field into the method, since its just used with a parent fingerprint. You can also match it to the parent fingerprints classDef

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
oSumAtrIX, line 31:
```suggestion
    val lyricsProviderUrlOption by stringOption(
```

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
```diff
@@ -0,0 +1,133 @@
...
+                val label = method.getInstruction<BuilderInstruction21t>(ifValidUrlIndex).target
+
+                val (stringRegister, compareToRegister) = method.getInstruction<FiveRegisterInstruction>(urlComparisonIndex).let { it.registerC to it.registerD }
+
+                method.addInstructions(ifValidUrlIndex + 1,
+                    """
+                    const-string v$stringRegister, "$lyricsProviderUrl"
+                    
+                    invoke-virtual {v$compareToRegister, v$stringRegister}, Ljava/lang/Object;->equals(Ljava/lang/Object;)Z
+
```
oSumAtrIX, line 67:
```suggestion
                    invoke-virtual { v$compareToRegister, v$stringRegister }, Ljava/lang/Object;->equals(Ljava/lang/Object;)Z
```

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
```diff
@@ -0,0 +1,133 @@
...
+val customLyricsPatch = bytecodePatch(
+    name = "Custom lyrics provider",
+    description = "Make Spotify use a custom lyrics provider.",
+) {
+    compatibleWith("com.spotify.music")
+
+    val lyricsProviderUrlOption = stringOption(
+        key = "lyricsProviderUrl",
+        title = "Lyrics provider URL",
+        description = "The custom lyrics provider URL.",
```
oSumAtrIX, line 34: Might drop an example here?
```suggestion
        description = "The URL to a custom lyrics provider such as SpotifyMobileLyricsAPI: https://github.com/Natoune/SpotifyMobileLyricsAPI.",
```

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
oSumAtrIX: I guess then the option can be named lyricsUrlHost? lyricsHost could be confusing. Alternatively, the user can input any URL and the patch can parse it as an URL and extract the host for sanity purposes.

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
```diff
@@ -0,0 +1,133 @@
...
+                val ifValidUrlIndex = method.indexOfFirstInstructionOrThrow(
+                    urlComparisonIndex,
+                    Opcode.IF_NEZ
+                )
+
+                val label = method.getInstruction<BuilderInstruction21t>(ifValidUrlIndex).target
+
+                val (stringRegister, compareToRegister) = method.getInstruction<FiveRegisterInstruction>(urlComparisonIndex).let { it.registerC to it.registerD }
+
+                method.addInstructions(ifValidUrlIndex + 1,
```
oSumAtrIX, line 62:
```suggestion
                method.addInstructions(
                	ifValidUrlIndex + 1,
```

oSumAtrIX, line 62: A lot of lines with `method.`. Should just use method.apply { ... } since we patch a method.

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
```diff
@@ -0,0 +1,133 @@
...
+                    const-string v$stringRegister, "$lyricsProviderUrl"
+                    
+                    invoke-virtual {v$compareToRegister, v$stringRegister}, Ljava/lang/Object;->equals(Ljava/lang/Object;)Z
+                    
+                    move-result v$stringRegister
+                """
+                )
+
+                method.addInstruction(ifValidUrlIndex + 1 + 3,
+                    BuilderInstruction21t(Opcode.IF_NEZ, stringRegister, label))
```
oSumAtrIX, line 73: The magic constants need to be named/explained properly.
```suggestion
                method.addInstruction(
                    ifValidUrlIndex + 1 + 3,
                    BuilderInstruction21t(Opcode.IF_NEZ, stringRegister, label)
                )
```

oSumAtrIX (changes requested):
Upon skimming the patch I am not sure whats going on in a fair amount of time. The complexity needs to be addressed. Organize the code into blocks of related code and add // region comments and comments explaining the strategy, magic constants etc

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
```diff
@@ -0,0 +1,133 @@
...
+            classDef.methods.add(
+                patchedClientMethod!!.toMutable().apply {
+                    addInstructions(method.instructions)
+
+                    replaceInstruction(clientStringAssignmentIndex, "const-string v$urlRegister, \"$lyricsProviderUrl\"")
+                }
+            )
+        }
+
+        with(executeFingerprint.method) {
```
oSumAtrIX, line 110: apply should be used instead of with as we apply changes to the method

Brosssh:
After more tests, I noticed that 2 changes were not needed for the patch to work, so they have been removed. I have added comments both in the fingerprint file and in the execute file, explaining what is being done. Most of the comments above have been resolved, I've left open the one that to me are still to clarify.

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/Fingerprints.kt
Brosssh: Sorry not clear what you mean here.

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
Brosssh: I opted for something easy but that should work good here. Let me know if it's ok.

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
```diff
@@ -0,0 +1,108 @@
...
+    name = "Change lyrics provider",
+    description = "Changes the lyrics provider to a custom one.",
+) {
+    compatibleWith("com.spotify.music")
+
+    var lyricsUrlHost by stringOption(
+        key = "lyricsProviderUrl",
+        title = "Lyrics provider URL/hostname",
+        description = "The URL or the hostname to a custom lyrics provider, " +
+                "such as SpotifyMobileLyricsAPI: https://github.com/Natoune/SpotifyMobileLyricsAPI.",
```
oSumAtrIX, line 30: I guess if a default option value is specified the specific mention of it can be removed.

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
```diff
@@ -0,0 +1,108 @@
...
+                    replaceInstruction(
+                        stringMatches!!.first().index,
+                        "const-string v$urlRegister, \"$lyricsUrlHost\"")
+                }
+            )
+        }
+
+        /**
+         * Adjust the lyrics HTTP builder method name to the method defined above.
+         * In this way the copied method is called rather than the stock one, returning the patched HTTP builder.
```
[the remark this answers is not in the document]
Brosssh: Correct, the original method is the http builder which is used like everywhere else, for every other non-modified request. I'll add a comment.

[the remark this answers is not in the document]
oSumAtrIX: Can't you add an if check and replace the host if it is the original spotfiy lyrics host?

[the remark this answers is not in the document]
Brosssh: Adding a few JADX screenshot to clarify. 

The function that returns the HTTP client:
image
This client is used for a lot of requests, so the host cannot be replaced here.

The method we need to patch:
image


How would I be able to do what you're saying? I need the client object, hence the duplication of `getHttpClient` with our custom host.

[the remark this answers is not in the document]
oSumAtrIX: Maybe the if check can be done on the path passed to the builder. So if the path is for lyrics, then replace the host.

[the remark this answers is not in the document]
oSumAtrIX: Alternatively, an instance of the client can possibly be turned back to a builder if its for lyrics by stubbing the APIs in an extension.

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
```diff
@@ -0,0 +1,133 @@
...
+        key = "lyricsProviderUrl",
+        title = "Lyrics provider URL/hostname",
+        description = "The URL or the hostname to a custom lyrics provider.",
+        required = true,
+        default = "lyrics.natanchiodi.fr"
+    ) {
+        val host = it!!
+            .removePrefix("http://")
+            .removePrefix("https://")
+            .substringBefore("/")
```
oSumAtrIX, line 40: An URL parser should be used, no string manipulation

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
```diff
@@ -0,0 +1,133 @@
...
+
+        try {
+            InetAddress.getByName(host)
```
oSumAtrIX, line 43: This wont work on IP addresses

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
```diff
@@ -0,0 +1,133 @@
...
+
+@Suppress("unused")
+val customLyricsPatch = bytecodePatch(
+    name = "Change lyrics provider",
+    description = "Changes the lyrics provider to a custom one.",
+) {
+    compatibleWith("com.spotify.music")
+
+    val lyricsUrlHost by stringOption(
+        key = "lyricsProviderUrl",
```
oSumAtrIX, line 31: Key should be the variable name. In the future key will be removed and instead the variable name will be used (or the title)

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
```diff
@@ -0,0 +1,133 @@
...
+        title = "Lyrics provider URL/hostname",
```
oSumAtrIX, line 32: Just the "URL" is enough. A URL is at least the hostname/IP. The patch can sanitize and extract the relevant part of the URL for further use.

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
```diff
@@ -0,0 +1,133 @@
...
+                "Host $host did not resolve to any domain."
+            )
+        }
+        true
+    }
+
+    execute {
+        var patchedClientMethod: ImmutableMethod?
+
+        clientBuilderFingerprint.apply {
```
oSumAtrIX, line 55: Fingerprints are immutable. Apply is incorrect semantics. Simply use the fingerprint wherever you need it

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
```diff
@@ -0,0 +1,133 @@
...
+            /**
+             * Copy the method definition of the HTTP client builder for a valid hostname.
+             */
+            patchedClientMethod = ImmutableMethod(
```
oSumAtrIX, line 59: Use // region comments instead of multiline

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
```diff
@@ -0,0 +1,133 @@
...
+                method.definingClass,
+                "getCustomLyricsProviderClient",
```
oSumAtrIX, line 61: Consistent with the name of the original method
```suggestion
                "getCustomLyricsProviderHttpClient",
```

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
oSumAtrIX, line 59: You dont need to construct a method like this. To create a copy it is enough to use .toMutable() on the original method and replace the implementation and name with your own.

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
```diff
@@ -0,0 +1,133 @@
...
+                method.parameters,
+                method.returnType,
+                method.accessFlags,
+                null,
+                null,
+                MutableMethodImplementation(6),
```
oSumAtrIX, line 67: Dont use magic constants. If the original method had 6, reference registerCount from there

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
```diff
@@ -0,0 +1,133 @@
...
+             * Copy the instructions into the new method and replace the assigned hostname with our host.
+             */
+            classDef.methods.add(
+                patchedClientMethod!!.toMutable().apply {
+                    addInstructions(method.instructions)
+
+                    replaceInstruction(
+                        urlAssignmentIndex,
+                        "const-string v$urlRegister, \"$lyricsUrlHost\"")
+                }
```
oSumAtrIX, line 86: This code should be moved to where the mutable implementation is constructed with an .apply

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
```diff
@@ -0,0 +1,133 @@
...
+            returns(clientBuilderFingerprint.originalMethod.returnType)
+            parameters()
+            opcodes(Opcode.CHECK_CAST)
+        }
+
+        /**
+         * Adjust the lyrics HTTP builder method name to the method defined above.
+         * In this way the copied method is called rather than the stock one, returning the patched HTTP builder.
+         */
+        lyricsHttpClientDefinitionFingerprint.method.apply {
```
oSumAtrIX, line 104: you should probably match it to a fingerprint class/method:

lyricsHttpClientDefinitionFingerprint.match(method or class).method

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
```diff
@@ -0,0 +1,133 @@
...
+                BuilderInstruction35c(
+                    Opcode.INVOKE_STATIC,
+                    getLyricsClientInstruction.registerCount,
+                    getLyricsClientInstruction.registerC,
+                    getLyricsClientInstruction.registerD,
+                    getLyricsClientInstruction.registerE,
+                    getLyricsClientInstruction.registerF,
+                    getLyricsClientInstruction.registerG,
+                    ImmutableMethodReference(
+                        patchedClientMethod!!.definingClass,
```
oSumAtrIX, line 124: Using assert not null everywhere where patchedClientMethod is used is not elegant. If you get rid of the apply above on the fingerprint you can inline the variable to what you assign it to and get rid of the nullability from the type

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/Fingerprints.kt
```diff
@@ -0,0 +1,11 @@
+package app.revanced.patches.spotify.misc.lyrics
+
+import app.revanced.patcher.fingerprint
+
+/**
+ * This method builds and returns an HTTP client for a hardcoded host (declared in this method).
+ */
+internal val clientBuilderFingerprint = fingerprint {
```
oSumAtrIX, line 8: No reason to use multline
```suggestion
// Method builds and returns an HTTP client for a hardcoded host (declared in this method).
internal val clientBuilderFingerprint = fingerprint {
```

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
```diff
@@ -0,0 +1,133 @@
...
+import com.android.tools.smali.dexlib2.iface.instruction.OneRegisterInstruction
+import com.android.tools.smali.dexlib2.immutable.ImmutableMethod
+import com.android.tools.smali.dexlib2.immutable.reference.ImmutableMethodReference
+import java.net.InetAddress
+import java.net.UnknownHostException
+import java.util.logging.Logger
+
+
+@Suppress("unused")
+val customLyricsPatch = bytecodePatch(
```
oSumAtrIX, line 24: The variable should be named according to the patch name or vice versa. Also don't forget to rename the file

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
Brosssh: What should be the correct way?

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
Brosssh: Could I replace the `apply` with a `with(clientBuilderFingerprint)`?

oSumAtrIX: You shouldnt need the fingerprint more than a coupe of times. No need of with() needed. Also https://github.com/ReVanced/revanced-patches/pull/4937#discussion_r2084328585

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
Brosssh: The original method name is `a` :D, anyway I have adjusted it with this.

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
oSumAtrIX: Or construct an URL object and validate that way

oSumAtrIX: That may be the case however this approach still relies on direct string manipulation operations and that should not happen under any circumstance. The current approach already shows fragility in the example of using uppercase letters for "HTTPS" which is a direct example of why you shouldn't do direct string manipulations. If the option asks for an URL, the option should assert it is an URL by specification and that is done by creating an URL object the easiest.

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
oSumAtrIX: This is fine. Precautious sanitization like this is fine. This avoids issues with users adding slashes, omitting protocols, using IP addresses or hostnames and issues of other likes.

Brosssh:
This should also support legacy version right? Right now it's not working.

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
Brosssh: `val host = URI(it!!).host` doesn't work if the user provide the hostname (without `https://`), tho.

oSumAtrIX: First construct URL, checking if it's an URL, if throws, create a dummy URL, but attempt setting the host to the option value. Then obtain the host from it. If that fails too, its also not a valid host

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
oSumAtrIX: The main issue is how the URL.is validated, not how you check if the host resolves

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
Brosssh: Can't `val host = URI(it!!).host ?: it` be sufficient? In case only the hostname is provided, it will not raise but it will return `null`, hence the `?:`.

I tested with:
- `lyrics.natanchiodi.fr` ->`lyrics.natanchiodi.fr`
- `https://lyrics.natanchiodi.fr/` -> `lyrics.natanchiodi.fr`
- `e` (random string to simulate error) -> `e` -> ` Host "e" did not resolve to any domain.` 
- `127.0.0.1`  -> `127.0.0.1`
- `http://127.0.0.1`  -> `127.0.0.1`

Seems to be working good.

oSumAtrIX: Does URL not throw an error? If not, it can still be malformed. Imagine a leading whitespace or slash

Brosssh: `URI(it!!).host` raise an `URISyntaxException` if the URL contains spaces (for example). I have added a try catch for this, and if the URL is malformed the validator returns false with a severe log.

oSumAtrIX: You don't have to log yourself, the frontends catch a validation exception and log respectively

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/Fingerprints.kt
```diff
@@ -0,0 +1,11 @@
...
+    strings("spclient.wg.spotify.com")
```
JingMatrix, line 9: `string("client == null")` could be an alternative fingerprint.

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
Brosssh: Though a custom error would be more user friendly but sure, removed.

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/ChangeLyricsProviderPatch.kt
```diff
@@ -0,0 +1,126 @@
...
+        )
+
+        //endregion
+
+        /**
+         * This method is where the HTTP client for lyrics is defined.
+         * This patch will replace this HTTP client with a patched HTTP client for the required custom lyrics host.
+         */
+        val lyricsHttpClientDefinitionFingerprint = fingerprint {
+            returns(clientBuilderFingerprint.originalMethod.returnType)
```
JingMatrix, line 90: This fingerprint mathes multiple methods, am I correct ?

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/Fingerprints.kt
Brosssh: `"client == null"` is the error raised string which could get removed/altered at any time in the app without changing it's behaviour. I've opted not to use it.

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/ChangeLyricsProviderPatch.kt
Brosssh: Unique to me, what app version are you using?

JingMatrix: I am using 9.0.42.529, the latest one in Play Store.
For me, `clientBuilderFingerprint.originalMethod.returnType` is `Lp/aka0`,
Then, this fingerprint gives `Lp/zja0;->e()Lp/aka0;` and `Lp/bff;->r4()Lp/aka0`.

I guess we use the method called by `com.spotify.connectivity.httpretrofit.RetrofitMaker`, am I correct?

Brosssh: Are you using `opcodes(Opcode.CHECK_CAST)`? Anyway, fingerprint will be improved with this https://github.com/ReVanced/revanced-patches/pull/4937#discussion_r2085480824

JingMatrix: You are right, after adding `opcodes(Opcode.CHECK_CAST)`, the method is unique.

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/CustomLyricsPatch.kt
Brosssh: This seems to be working.. Thoughts?
image

Brosssh: Sorry, forgot to say, I have renamed it. The method name is indeed obfuscated.

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/ChangeLyricsProviderPatch.kt
```diff
@@ -0,0 +1,148 @@
...
+    description = "Changes the lyrics provider to a custom one.",
+) {
+    compatibleWith("com.spotify.music")
+
+    val lyricsProviderUrl by stringOption(
+        key = "lyricsProviderUrl",
+        title = "Lyrics provider URL",
+        description = "The URL to a custom lyrics provider.",
+        required = true,
+        default = "lyrics.natanchiodi.fr"
```
oSumAtrIX, line 38: Since the option asks for an URL
```suggestion
        default = "https://lyrics.natanchiodi.fr"
```

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/ChangeLyricsProviderPatch.kt
```diff
@@ -0,0 +1,148 @@
...
+    ) {
+        val host: String?
+        try {
+            host = URI(it!!).host ?: it
+        } catch (e: URISyntaxException) {
+            return@stringOption false
+        }
```
oSumAtrIX, line 45:
```suggestion
        val host = try {
            URI(it!!).host ?: it
        } catch (e: URISyntaxException) {
            return@stringOption false
        }
```

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/ChangeLyricsProviderPatch.kt
```diff
@@ -0,0 +1,148 @@
...
+
+        try {
+            InetAddress.getByName(host)
+        } catch (e: UnknownHostException) {
+            Logger.getLogger(this::class.java.name).warning(
+                "Host \"$host\" did not resolve to any domain."
+            )
```
oSumAtrIX, line 51: This is just a DNS check. I don't think it's the patch's objective to check for DNS issues. Reachability has to be checked differently, not just via DNS because DNS is only for domain names. The reachability check is ineffective for IP addresses.

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/ChangeLyricsProviderPatch.kt
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    execute {
+        if(IS_SPOTIFY_LEGACY_APP_TARGET) {
+            Logger.getLogger(this::class.java.name).severe(
+                "Change lyrics provider patch is not supported for this target version."
+            )
+            return@execute
+        }
+
```
oSumAtrIX, line 63:
```suggestion
```

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/ChangeLyricsProviderPatch.kt
```diff
@@ -0,0 +1,148 @@
...
+
+        //region Create a patched HTTP client for the requested URL
```
oSumAtrIX, line 65:
```suggestion
        //region Create a patched HTTP client for the requested URL.
```

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/ChangeLyricsProviderPatch.kt
oSumAtrIX: The patch isn't specific to this repository

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/ChangeLyricsProviderPatch.kt
```diff
@@ -0,0 +1,148 @@
...
+                urlAssignmentIndex,
+            ).registerA
+
+            patchedClientMethod = method.toMutable().apply {
+                name = "patch_getCustomLyricsProviderHttpClient"
+
+                replaceInstruction(
+                    urlAssignmentIndex,
+                    "const-string v$urlRegister, \"$lyricsProviderUrl\""
+                )
```
oSumAtrIX, line 82: toMutable() is not copying the implementation. You are modifying the original instructions. Was the patch tested?

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/ChangeLyricsProviderPatch.kt
```diff
@@ -0,0 +1,148 @@
...
+                "Change lyrics provider patch is not supported for this target version."
+            )
+            return@execute
+        }
+
+
+        //region Create a patched HTTP client for the requested URL
+
+        // Copy the method definition of the HTTP client builder for a valid hostname.
+        val patchedClientMethod : MutableMethod
```
oSumAtrIX, line 67: Naming consistency is important
```suggestion
        val patchedHttpClientBuilderMethod : MutableMethod
```

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/ChangeLyricsProviderPatch.kt
```diff
@@ -0,0 +1,148 @@
...
+
+        clientBuilderFingerprint.apply {
```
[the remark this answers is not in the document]
oSumAtrIX: Apply suggests semantics that are not valid in this context

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/ChangeLyricsProviderPatch.kt
```diff
@@ -0,0 +1,146 @@
...
+                replaceInstruction(
+                    urlAssignmentIndex,
+                    "const-string v$urlRegister, \"$lyricsProviderUrl\""
+                )
+            }
+
+            classDef.methods.add(
+                patchedClientMethod
+            )
+        }
```
oSumAtrIX, line 88:
```suggestion
        val patchedClientMethod = clientBuilderFingerprint.apply {
            val urlAssignmentIndex = stringMatches!!.first().index

            val urlRegister = method.getInstruction<OneRegisterInstruction>(
                urlAssignmentIndex,
            ).registerA

            method.toMutable().apply {
                name = "patch_getCustomLyricsProviderHttpClient"

                replaceInstruction(
                    urlAssignmentIndex,
                    "const-string v$urlRegister, \"$lyricsProviderUrl\""
                ).also(classDef.methods::add)
            }
        }
```

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/ChangeLyricsProviderPatch.kt
```diff
@@ -0,0 +1,146 @@
...
+            returns(clientBuilderFingerprint.originalMethod.returnType)
+            parameters()
+            custom { method, _ ->
+                method.indexOfFirstInstruction {
+                    getReference<MethodReference>() == clientBuilderFingerprint.originalMethod
+                } >= 0
+            }
+        }
+
+        //region Use the patched HTTP client for lyrics request
```
oSumAtrIX, line 109:
```suggestion
        // region Conditionally use the modified HTTP client builder method when the client is used for the lyrics API.
```

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/ChangeLyricsProviderPatch.kt
oSumAtrIX: A domain is valid if it does not resolve though. I might set up the DNS at a later stage. It is not the patches business to enforce it being done before.

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/ChangeLyricsProviderPatch.kt
oSumAtrIX: > Any protocol entered here is ignored.

Yes, however the option asks for an URL and a URL asks for a protocol. 

> What if the user enters something stupid like:
> ftp://lyrics.natanchiodi.fr
> The patch will happily accept and use it as-if nothing is wrong.

Nothing is wrong. The option asked for an URL, the user entered one. The host is extracted from it.

> Or what if they thing they can use a local server without https

The lyrics backend setups https. The user can enter HTTP or https regardless of that for the reason.
The option asks for an URL so users might assume the protocol will be respected. You can add to the description, that the host only matters.

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/ChangeLyricsProviderPatch.kt
```diff
@@ -0,0 +1,146 @@
...
+        if(IS_SPOTIFY_LEGACY_APP_TARGET) {
+            Logger.getLogger(this::class.java.name).severe(
+                "Change lyrics provider patch is not supported for this target version."
+            )
+            return@execute
+        }
+
+        // region Create a modified copy of the HTTP client builder method with the custom lyrics provider URL.
+
+        // Copy the method definition of the HTTP client builder for a valid hostname.
```
oSumAtrIX, line 66: redundant due to the region comment
```suggestion
```

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/ChangeLyricsProviderPatch.kt
Brosssh: Yes. Double checked now and it's duplicating the method:
image

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/ChangeLyricsProviderPatch.kt
```diff
@@ -0,0 +1,146 @@
...
+
+        //endregion
+
+        /**
+         * The method where the HTTP client for lyrics is defined.
+         *
+         * This patch will replace this HTTP client with a patched HTTP client for the required
+         * custom lyrics host. This new method is only used for the lyrics request and nowhere else.
+         */
+        val lyricsHttpClientDefinitionFingerprint = fingerprint {
```
oSumAtrIX, line 89:
```suggestion
        val buildLyricsHttpClientFingerprint = fingerprint {
```

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/ChangeLyricsProviderPatch.kt
```diff
@@ -0,0 +1,146 @@
...
+                method.indexOfFirstInstruction {
+                    getReference<MethodReference>() == httpClientBuilderFingerprint.originalMethod
+                } >= 0
+            }
+        }
+
+        // region Conditionally use the modified HTTP client builder method when the client is used for the lyrics API.
+
+        lyricsHttpClientDefinitionFingerprint.method.apply {
+            val getLyricsClientIndex = indexOfFirstInstructionOrThrow() {
```
oSumAtrIX, line 102:
```suggestion
        }.method.apply {
            val getLyricsClientIndex = indexOfFirstInstructionOrThrow() {
```

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/Fingerprints.kt
```diff
@@ -0,0 +1,16 @@
+package app.revanced.patches.spotify.misc.lyrics
+
+import app.revanced.patcher.fingerprint
+
+/**
+ * This method builds and returns an HTTP client for a hardcoded host (declared in this method).
+ */
+internal val httpClientBuilderFingerprint = fingerprint {
```
oSumAtrIX, line 8: Not sure why multiline is used here

```suggestion
// This method builds and returns an HTTP client with a hardcoded host (declared in this method).
internal val httpClientBuilderFingerprint = fingerprint {
```

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/ChangeLyricsProviderPatch.kt
```diff
@@ -0,0 +1,127 @@
...
+                    urlAssignmentIndex,
+                    "const-string v$urlRegister, \"$lyricsProviderHost\""
+                )
+                classDef.methods.add(this)
+            }
+        }
+
+        //endregion
+
+        // Replace the call to the HTTP client builder method used exclusively for lyrics by the modified one.
```
drobotk, line 88: should say `region` here, for the `endregion` in L125

Brosssh:
Should this has `use = false` btw? It shouldn't be on as default.

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/ChangeLyricsProviderPatch.kt
```diff
@@ -0,0 +1,133 @@
...
+                getReference<MethodReference>()?.returnType == "Lokhttp3/HttpUrl;"
+            }
+
+            val urlRegister = method
+                .getInstruction<FiveRegisterInstruction>(urlBuilderIndex - 1)
+                .registerD
+
+            MutableMethod(method).apply {
+                name = "patch_getCustomLyricsProviderHttpClient"
+                addInstruction(
```
oSumAtrIX, line 84: This will add to the original method as well. The implementation is copied by reference

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/ChangeLyricsProviderPatch.kt
```diff
@@ -0,0 +1,133 @@
...
+        } catch (e: UnknownHostException) {
+            Logger.getLogger(this::class.java.name).warning(
+                "Host \"$host\" did not resolve to any domain."
+            )
+        }
+        true
+    }
+
+    execute {
+        if(IS_SPOTIFY_LEGACY_APP_TARGET) {
```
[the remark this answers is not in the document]
oSumAtrIX: That's one example that would solve the issue here, works

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/ChangeLyricsProviderPatch.kt
Brosssh: The method is correctly copied. The old one is the same.
image

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/ChangeLyricsProviderPatch.kt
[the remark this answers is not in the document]
Brosssh: Oh nevermind, do you mean if I deselect premium patch it won't work?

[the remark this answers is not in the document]
Brosssh: It works with only this patch tho..
image

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/ChangeLyricsProviderPatch.kt
```diff
@@ -0,0 +1,140 @@
...
+            MutableMethod(method).apply {
+                name = "patch_getCustomLyricsProviderHttpClient"
+                classDef.methods.add(this)
+            }.also {
+                method.addInstruction(
+                    urlBuilderIndex - 1,
+                    "const-string v$urlRegister, \"$lyricsProviderHost\""
+                )
+            }
+        }
```
oSumAtrIX, line 96:
```suggestion

            MutableMethod(method).apply {
                name = "patch_getCustomLyricsProviderHttpClient"
                addInstruction(
                    urlBuilderIndex - 1,
                    "const-string v$urlRegister, \"$lyricsProviderHost\""
                )
            }.apply(classDef.methods::add)
        }
```

patches/src/main/kotlin/app/revanced/patches/spotify/misc/lyrics/ChangeLyricsProviderPatch.kt
```diff
@@ -0,0 +1,140 @@
...
+            val urlRegister = method
+                .getInstruction<FiveRegisterInstruction>(urlBuilderIndex - 1)
+                .registerD
+
+            MutableMethod(method).apply {
+                name = "patch_getCustomLyricsProviderHttpClient"
+                classDef.methods.add(this)
+            }.also {
+                method.addInstruction(
+                    urlBuilderIndex - 1,
```
oSumAtrIX, line 92: This will still add the instruction to the original method because the implementation is copied by reference when you call MutableMethod()

drobotk:
Needs updating, now doesn't need to depend on sharedExtensionPatch

PR #4977, opened 2025-05-17
feat: Add `Hide app icon` patch

### LINKED ISSUE #3575: feat: (universal) - Add `Hide from Launcher` option for universal.
### Feature description

This patch will allow every app to get hidden from the launcher. As there are so many apps which are required but we don't want their app icon to be down in launcher. This patch will allow the app icon to get hidden from Launcher.

### Motivation

Custom ROM and Different UI can give permission to hide the app. But the stock Android users don't have that power, specially old Android version users like 10. This patch will be very helpful for those users.

### Acknowledgements

- [X] I have checked all open and closed feature requests and this is not a duplicate
- [X] I have chosen an appropriate title.
- [X] All requested information has been provided properly.

ILoveOpenSourceApplications:
Hides the app icon from the launcher.

After patching, the app can only be opened in a few ways.

#### App links
The app supports opening web links in app, and the app info 'Opens supported links' and all supported web links are enabled. App is then opened using a web link, or opened using Android assistant app ("Hey Google, open app Whatever").
&nbsp;

#### App is launched from another app
The app has a custom intent that other apps can launch from, such as MicroG that can be opened from ReVanced YouTube settings.  Apps with deep integration in other apps may also support this (such as opening Facebook Messenger from Facebook, or opening Amazon Kindle from the Amazon shopping app).

Closes #3575.

### DIFF
--- a/patches/src/main/kotlin/app/revanced/patches/all/misc/appicon/HideAppIconPatch.kt
+++ b/patches/src/main/kotlin/app/revanced/patches/all/misc/appicon/HideAppIconPatch.kt
@@ -2,7 +2,6 @@ package app.revanced.patches.all.misc.appicon
 
 import app.revanced.patcher.patch.resourcePatch
 import org.w3c.dom.Element
-import java.util.logging.Level
 import java.util.logging.Logger
 
 private const val ANDROID_NS = "http://schemas.android.com/apk/res/android"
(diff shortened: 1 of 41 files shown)


ILoveOpenSourceApplications:
@vippium, any idea why it's failing to hide the app icon?

ILoveOpenSourceApplications:
> Was this done using AI? The `// ** text ** ` comments resemble ChatGPT or DeepSeek.

Yes, actually.

vippium:
> If this can be finished, also verify this works with MicroG.

It works. I tried it with official rv microg.

vippium:
> If this can be finished, also verify this works with MicroG.

Yupp...this works with microg.

vippium:
> @vippium, any idea why it's failing to hide the app icon?

Which one ?

vippium:
> @vippium, any idea why it's failing to hide the app icon?

But, in my case, it is working with every app. Even with YouTube. And, for cross-check, I also tried it with Lawnchair, Niagara and Smart too.

vippium:
> @vippium are you using the latest code that's here?
> 
> When I tried patching YT it failed during patching and did not make any changes.

Ahh...It's very typical for me to use the CLI for checking the PR. Let me know if you can generate me an rvp file for it. So, that I can use it with manager.

oSumAtrIX:
For reference https://github.com/ReVanced/revanced-patches/actions/runs/15218048266/workflow#L34

vippium:
> But what's here does not work with YouTube. Possibly because of the [hard coded "launcher" name](https://github.com/ReVanced/revanced-patches/blob/dae785a71ca957d99b3380e762a9f54b0d6bc34e/patches/src/main/kotlin/app/revanced/patches/all/misc/appicon/HideAppIconPatch.kt#L37) it's looking for.

After checking it with the patch, I can comfirm that it is not working. But how then it works if we change the launcher activity manually, with YT.

Let me check if it works same way with others or not.

vippium:
> PR build artifacts are generated for a while now, see the workflow runs

Well, thnx oSumAtrIX for this.

vippium:
> That is a warning, and is only displayed if the patch fails to find the manifest entries it's expecting. It's showing that because the patch is broken and it's not making any changes.

Ohh... I thought something else.

ILoveOpenSourceApplications:
@LisoUseInAIKyrios, what about opening the app from `Settings > Apps > (AppPatchedWithHideAppIcon) > Open`. Does that work as well?

ILoveOpenSourceApplications:
Oh, okay. Then @vippium, can you check this?

> what about opening the app from `Settings > Apps > (AppPatchedWithHideAppIcon) > Open`. Does that work as well?

vippium:
> My device does not show the open action in the app info.

It won't show that option. It only shows for those apps which have the launcher activity.

If user is patching the app with this patch, I think user knows what he's doing.

vippium:
> Should the patch be modified to show `Open` in settings page by adding a new fake “alias” activity with the MAIN + LAUNCHER intent and then immediately disable that alias via `android:enabled="false"`?

You can go with it if you prefer this.

Instead of that, you can go with `Open settings in App` option in settings. And if you're going with both, then mark them as optional. As if user have alternate ways to access the app (like via qs), then he can directly skip that option. And if not, he can go with the option.

vippium:
> The latest change has broke the patch. It now shows up again in the launcher.

Great...

patches/src/main/kotlin/app/revanced/patches/all/misc/appicon/HideAppIconPatch.kt
```diff
@@ -0,0 +1,43 @@
...
+val hideAppIconPatch = resourcePatch(
+    name = "Hide app icon",
+    description = "Hides the app icon from the Android launcher.",
+    use = false,
+) {
+    execute {
+        document("AndroidManifest.xml").use { document ->
+            var changed = false
+
+            val intentFilters = document.getElementsByTagName("intent-filter")
```
oSumAtrIX, line 17: You can use the helper extension function "childElementSequence"

ILoveOpenSourceApplications: Yeah, it was missing an 's'.

PR #4998, opened 2025-05-19
Add patch to disable ads on cricbuzz app

agrawalsourav98:
Add patch to disable ads on cricbuzz app

### DIFF
--- a/patches/src/main/kotlin/app/revanced/patches/cricbuzz/DisableAdsPatch.kt
+++ b/patches/src/main/kotlin/app/revanced/patches/cricbuzz/DisableAdsPatch.kt
@@ -1,0 +1,33 @@
+package app.revanced.patches.cricbuzz
+
+import app.revanced.patcher.extensions.InstructionExtensions.addInstructions
+import app.revanced.patcher.extensions.InstructionExtensions.getInstruction
+import app.revanced.patcher.fingerprint
+import app.revanced.patcher.patch.bytecodePatch
+import com.android.tools.smali.dexlib2.Opcode
+import com.android.tools.smali.dexlib2.iface.instruction.formats.Instruction11x
+
+@Suppress("unused")
+val disableAdsPatch = bytecodePatch (
+    "Disable ads",
+) {
+    compatibleWith("com.cricbuzz.android")
+
+    execute {
+        // modify user state as "ACTIVE" during switch statement
+        userStateSwitchFingerprint.method.apply {
+            // find the line where the switch variable is assigned
+            // and replace the register with a constant string "ACTIVE"
+            val lineFingerprint = fingerprint {
+                opcodes(Opcode.MOVE_RESULT_OBJECT)
+            }
+            val insertIndex = lineFingerprint.match(userStateSwitchFingerprint.method).patternMatch!!.startIndex
+            val register = getInstruction<Instruction11x>(insertIndex).registerA
+
+            addInstructions(
+                insertIndex + 1,
+                "const-string v$register, \"ACTIVE\""
+            )
+        }
+    }
+}
--- a/patches/src/main/kotlin/app/revanced/patches/cricbuzz/Fingerprints.kt
+++ b/patches/src/main/kotlin/app/revanced/patches/cricbuzz/Fingerprints.kt
@@ -1,0 +1,9 @@
+package app.revanced.patches.cricbuzz
+
+import app.revanced.patcher.fingerprint
+import com.android.tools.smali.dexlib2.Opcode
+
+internal val userStateSwitchFingerprint = fingerprint {
+    strings("key.user.state","NA")
+    opcodes(Opcode.SPARSE_SWITCH)
+}


patches/src/main/kotlin/app/revanced/patches/cricbuzz/Fingerprints.kt
```diff
@@ -0,0 +1,9 @@
+package app.revanced.patches.cricbuzz
+
+import app.revanced.patcher.fingerprint
+import com.android.tools.smali.dexlib2.Opcode
+
+internal val userStateSwitchFingerprint = fingerprint {
+    strings("key.user.state","NA")
```
oSumAtrIX, line 7:
```suggestion
    strings("key.user.state", "NA")
```

patches/src/main/kotlin/app/revanced/patches/cricbuzz/DisableAdsPatch.kt
```diff
@@ -0,0 +1,33 @@
...
+val disableAdsPatch = bytecodePatch (
+    "Disable ads",
+) {
+    compatibleWith("com.cricbuzz.android")
+
+    execute {
+        // modify user state as "ACTIVE" during switch statement
+        userStateSwitchFingerprint.method.apply {
+            // find the line where the switch variable is assigned
+            // and replace the register with a constant string "ACTIVE"
```
oSumAtrIX, line 20: Code is self explanatory already
```suggestion
        userStateSwitchFingerprint.method.apply {
```

patches/src/main/kotlin/app/revanced/patches/cricbuzz/DisableAdsPatch.kt
```diff
@@ -0,0 +1,33 @@
...
+            val lineFingerprint = fingerprint {
+                opcodes(Opcode.MOVE_RESULT_OBJECT)
+            }
+            val insertIndex = lineFingerprint.match(userStateSwitchFingerprint.method).patternMatch!!.startIndex
+            val register = getInstruction<Instruction11x>(insertIndex).registerA
+
+            addInstructions(
+                insertIndex + 1,
```
oSumAtrIX, line 28: If `lineFingerprint.match(userStateSwitchFingerprint.method).patternMatch!!.startIndex` is `insertIndex` why are we inserting at insertIndex + 1. The variable should be called differently, or assign the variable to the startIndex + 1

patches/src/main/kotlin/app/revanced/patches/cricbuzz/DisableAdsPatch.kt
```diff
@@ -0,0 +1,33 @@
...
+    compatibleWith("com.cricbuzz.android")
+
+    execute {
+        // modify user state as "ACTIVE" during switch statement
+        userStateSwitchFingerprint.method.apply {
+            // find the line where the switch variable is assigned
+            // and replace the register with a constant string "ACTIVE"
+            val lineFingerprint = fingerprint {
+                opcodes(Opcode.MOVE_RESULT_OBJECT)
+            }
```
oSumAtrIX, line 23: You can use indexOfFirstInstructionOrThrow here

oSumAtrIX (commented):
The patch and fingerprint file should be moved to an "ads" package

patches/src/main/kotlin/app/revanced/patches/cricbuzz/Fingerprints.kt
oSumAtrIX, line 7: What is a "user state"? It doesn't sound to me like its designed just for ads?

agrawalsourav98: Basically the app has a premium tier where this variable stores the current subscription status of the logged in user. The code uses the value from this to decide whether to display ads or not, and unlocks some parts of the app available to premium users.

oSumAtrIX: Then this patch is not a "Disable ads" patch but an "Unlock premium" patch which goes against the contribution guidelines of this repository. @LisoUseInAIKyrios What do you say.

patches/src/main/kotlin/app/revanced/patches/cricbuzz/DisableAdsPatch.kt
```diff
@@ -0,0 +1,33 @@
...
+import app.revanced.patcher.fingerprint
+import app.revanced.patcher.patch.bytecodePatch
+import com.android.tools.smali.dexlib2.Opcode
+import com.android.tools.smali.dexlib2.iface.instruction.formats.Instruction11x
+
+@Suppress("unused")
+val disableAdsPatch = bytecodePatch (
+    "Disable ads",
+) {
+    compatibleWith("com.cricbuzz.android")
```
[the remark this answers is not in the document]
oSumAtrIX: Version constraints should not be added when the patch is simple (like replacing a deterministic string). Only when the patch is complex enough that it calls for a version constraint.

patches/src/main/kotlin/app/revanced/patches/cricbuzz/Fingerprints.kt
agrawalsourav98: Well the premium parts of the app don't work, as the api calls require an access token, and the user is simply redirected to the login page.

agrawalsourav98: Yeah. Let me try if I can hide only the ads.

agrawalsourav98: yeah, I checked and it's being used at 124 places, I think most of them are ad related but the line kinda gets blurred at a lot of places. In short this field, controls ads,controls if subscription options are shown (not shown when this flag is set). I have checked and rest everything is the more or less the same as the stock app, as api calls without token don't get through.

patches/src/main/kotlin/app/revanced/patches/cricbuzz/ads/DisableAdsPatch.kt
```diff
@@ -0,0 +1,28 @@
...
+import app.revanced.patcher.patch.bytecodePatch
+import app.revanced.util.indexOfFirstInstructionOrThrow
+import com.android.tools.smali.dexlib2.Opcode
+import com.android.tools.smali.dexlib2.iface.instruction.OneRegisterInstruction
+
+@Suppress("unused")
+val disableAdsPatch = bytecodePatch (
+    "Hide ads",
+) {
+    compatibleWith("com.cricbuzz.android" ("6.23.02"))
```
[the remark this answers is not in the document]
Nuckyz: But the suggestion is to remove the space, not add it

PR #5004, opened 2025-05-20
fix(YouTube - Hide ads): Hide new type of buttoned ad

### LINKED ISSUE #4993: feat(YouTube): Block new ad type
### Feature description

There is a new ad type that takes up the entire screen in the feed, so it should be blocked.

Here's a log of it

[log.txt](https://github.com/user-attachments/files/20277465/new-ad.txt)

### Motivation

It's an ad that should also be blocked.

### Acknowledgements

ILoveOpenSourceApplications:
Closes #4993.


ILoveOpenSourceApplications:
Also, couldn't we merge some of these filters having common terms like using `_button_layout` instead for all the litho having it?

ILoveOpenSourceApplications:
Can we use a generalAdsIdentifier then, like [this](https://github.com/inotia00/revanced-patches/blob/42cc8201f5dfcc05f785e578475c4e7e52c3ecfd/extensions/shared/src/main/java/app/revanced/extension/youtube/patches/components/AdsFilter.java#L39-#L78)?

ILoveOpenSourceApplications:
I guess it goes both ways, whilst it may hide other unwanted elements, it's also a way to future proof in the case of hiding ads as well.

PR #5029, opened 2025-05-25
fix: Remove parameter erroneously passed to vararg

1fexd:


Nuckyz:
A patched app using this patch would crash on startup without this fix. Now it works properly and hides the ADB status without crashing. Perhaps this PR should be renamed according to this?

Issue #4855, opened 2025-04-24
feat: Remove threads ads


KEVIN970712:
### Feature description

![Image](https://github.com/user-attachments/assets/acfb9487-dacd-4f4d-a183-d41246fae923)

![Image](https://github.com/user-attachments/assets/c52e2d83-657a-4de0-bb50-aae0b4c61718)

### Motivation

Yesterday, Threads advertising business was announced to expand to the world, and the ads appeared in my country.

### Acknowledgements

- [x] I have checked all open and closed feature requests and this is not a duplicate
- [x] I have chosen an appropriate title.
- [x] All requested information has been provided properly.

PR #5064, opened 2025-05-29
feat(Threads): Hide Ads

scrazzz:
title


scrazzz:
Opened due to accidental closure of PR #5035 and fixes issue #4855

patches/src/main/kotlin/app/revanced/patches/meta/ads/HideAdsPatch.kt
```diff
@@ -0,0 +1,21 @@
+package app.revanced.patches.meta.ads
+
+import app.revanced.patcher.patch.bytecodePatch
+import app.revanced.util.returnEarly
+
+@Suppress("unused")
+val hideAdsPatch = bytecodePatch(
+    name = "Hide ads",
```
drobotk, line 8: Add a `description`.

drobotk (changes requested):

patches/src/main/kotlin/app/revanced/patches/meta/ads/HideAdsPatch.kt
drobotk: My bad.

PR #5081, opened 2025-05-31
fix(YouTube): Support A/B Shorts layout for RYD and component hiding

### LINKED ISSUE #5024: bug: Broken YouTube Shorts Player UI & Dislikes For Shorts Not Showing
### Bug description

For the last week or so, specifically using my main default account, ReVanced does not let me customize the Shorts Player UI whats so ever. I cannot remove buttons (i.e the 'Remix' button), and I cannot see the number of 'Dislikes' per Short. 

In fact the Shorts Player UI on the Revanced app looked practically identical to the way the Shorts Player UI looks on the regular YouTube app, because regardless of what settings I tweak within Revanced settings under the 'Shorts' and 'Return YouTube Dislikes' categories, nothing actually applies. 

This is happening across both my Smartphone and Tablet with the same account login.

I have tried updating and re-patching both the ReVanced Manager itself, and GmsCore/microG to the latest versions sourced from GitHub.

ReVanced Manager - https://github.com/ReVanced/revanced-manager/releases/tag/v1.24.0
GmsCore/microG - https://github.com/ReVanced/GmsCore/releases/tag/v0.3.1.4.240913

I have also tried updating to the latest recommended version of the YouTube APK for ReVanced as well (which is the 20.12.46) version). 

I've attempted to spoof the app version to various other older versions under 'General' pn both devices with no success.

It doesn't seem to be the actual 'Return YouTube Dislikes' API itself, because on Firefox, by installing the 'Return YouTube Dislikes' extension and then using it for the Shorts Player on the YouTube website with the same account login credentials on my Desktop PC, Smartphone and Tablet, does in fact allow me to see the number of Dislikes for Shorts directly within the Shorts Player.  

 
If I switch to a different account or use incognito mode, I can see the 'Dislikes' for Shorts in the Shorts Player again, but the moment I switch back to my main default account, the same thing happens. 

The 'Dislike' counter on ReVanced is still working perfectly fine for regular videos on the standard player. 

Image

Image

### Error logs

```shell

```

### Solution

_No response_

### Additional context

_No response_

### Acknowledgements

aliernfrog:
Adds support for a new Shorts layout currently being A/B tested. Resolves #5024.

RYD works fine, component hiding also works fine. However I couldn't test **stickers** and **like fountain** toggles as I could not get them to show.

It looks like YouTube also turned all the buttons in "All other action buttons" into `reel_action_button.eml`, so now they can only be hidden using their icons. Should the filters for those be shortened in case YouTube renames the icons in the future?

### DIFF
--- a/extensions/youtube/src/main/java/app/revanced/extension/youtube/patches/ReturnYouTubeDislikePatch.java
+++ b/extensions/youtube/src/main/java/app/revanced/extension/youtube/patches/ReturnYouTubeDislikePatch.java
@@ -152,11 +152,15 @@ public class ReturnYouTubeDislikePatch {
                 return original; // No need to check for Shorts in the context.
             }
 
-            if (conversionContextString.contains("|shorts_dislike_button.eml")) {
+            if (Utils.containsAny(conversionContextString,
+                    "|shorts_dislike_button.eml", "|reel_dislike_button.eml"
+            )) {
                 return getShortsSpan(original, true);
             }
 
-            if (conversionContextString.contains("|shorts_like_button.eml")) {
+            if (Utils.containsAny(conversionContextString,
+                    "|shorts_like_button.eml", "|reel_like_button.eml"
+            )) {
                 if (!Utils.containsNumber(original)) {
                     Logger.printDebug(() -> "Replacing hidden likes count");
                     return getShortsSpan(original, false);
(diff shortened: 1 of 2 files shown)


sliceanddice9:
So if I understand things correctly; the Shorts Player UI for Shorts Dislikes and Button Hiding Customization, has been fixed for those specific affected accounts (like mine) who were targeted with the A/B testing by YouTube? 

Not to rush you guys because I appreciate all the hard work you guys do but is there a estimate time frame on when the patch update will be available for installation?

Issue #5111, opened 2025-06-05
bug(Spotify): Hide Create button patch removes bottom shelf


NMinker:
### Bug description

After patching the new patches (5.26.0), Spotify removes the bottom shelf shelf when the "Hide Create button" patch is enabled.

The bottom shelf returns when the "Hide Create button" patch is disabled

![Image](https://github.com/user-attachments/assets/db0fadc7-eeb1-41c0-840d-2c2be0ac85d2)

### Error logs

```shell

```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] I have chosen an appropriate title.
- [x] All requested information has been provided properly.

Vitaspiros:
Having the same issue.

Nuckyz:
@LisoUseInAIKyrios the issue or hiding all buttons has been fixed, but the patch is still broken. I added some debug logs and it seems like getIdentifier is always returning 0, even though the resource exists. Happens in an app merged with antiplit m.

I have no idea why this is the case.

![Image](https://github.com/user-attachments/assets/cc9a8b92-53ed-49bb-975e-583b84694cdb)

![Image](https://github.com/user-attachments/assets/25e2472c-2a4b-43e1-b6df-a977f8a0021a)

Nuckyz:
> Can the old approach be used, and turn off client side feature flags?

Yes, but that's not how we want to continue. The patch is meant to avoid using experiments to hide the button. It works fine when the resource is returned properly, but in this case, it's just not working. Unknown if it could be related to Antisplit M. I also haven't found a reason getIdentifier could be failing like this when doing research

Nuckyz:
It seems to be an issue caused by antisplit m, using an apk from apkmirror and just having the Hide Create button patch has no issues. Also, adding a resource patch and patching the antisplit m apk fixes it not working, because resources are compiled again.

PR #5122, opened 2025-06-05
fix(Spotify): Prevent hiding all navigation bar buttons

Nuckyz:
Just now realized this issue. I saw some users claiming all navigation bar items were getting hidden, and this might be why.


ILoveOpenSourceApplications:
#5111 can be linked to this.

PR #5153, opened 2025-06-09
feat(Messenger - Remove Meta AI): Improve patch logic

drobotk:
The patch as it is now can have side effects (disabling non-Meta AI features) because its ID matching logic is too loose, because it accounts for the IDs changing across app versions. Now it targets only Meta AI config flags, by extracting the starting digits specific to the app version from code. Also added logging of the overriden flags, for potential future debugging.

TODO: write some comments explaining what's going on

### DIFF
--- a/extensions/messenger/src/main/java/app/revanced/extension/messenger/metaai/RemoveMetaAIPatch.java
+++ b/extensions/messenger/src/main/java/app/revanced/extension/messenger/metaai/RemoveMetaAIPatch.java
@@ -1,14 +1,21 @@
 package app.revanced.extension.messenger.metaai;
 
+import java.util.HashSet;
+import java.util.Set;
+
+import app.revanced.extension.shared.Logger;
+
 @SuppressWarnings("unused")
 public class RemoveMetaAIPatch {
+    private static final Set<Long> loggedIDs = new HashSet<>();
+
     public static boolean overrideBooleanFlag(long id, boolean value) {
-        // This catches all flag IDs related to Meta AI.
-        // The IDs change slightly with every update,
-        // so to work around this, IDs from different versions were compared
-        // to find what they have in common, which turned out to be those first bits.
-        // TODO: Find the specific flags that we care about and patch the code they control instead.
-        if ((id & 0x7FFFFFC000000000L) == 0x810A8000000000L) {
+        if (Long.toString(id).startsWith("REPLACED_BY_PATCH")) {
+            synchronized (loggedIDs) {
+                if (loggedIDs.add(id)) {
+                    Logger.printInfo(() -> "Overriding " + id + " from " + value + " to false");
+                }
+            }
             return false;
         }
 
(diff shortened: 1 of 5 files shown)


patches/src/main/kotlin/app/revanced/patches/messenger/metaai/Fingerprints.kt
```diff
@@ -11,3 +14,21 @@ internal val getMobileConfigBoolFingerprint = fingerprint {
         classDef.interfaces.contains("Lcom/facebook/mobileconfig/factory/MobileConfigUnsafeContext;")
     }
 }
+
+internal val relevantIDContainingMethodFingerprint = fingerprint {
```
Nuckyz, line 18: The fingerprint name should reflect what happens in the code

patches/src/main/kotlin/app/revanced/patches/messenger/misc/extension/Hooks.kt
```diff
@@ -2,6 +2,8 @@ package app.revanced.patches.messenger.misc.extension
 
 import app.revanced.patches.shared.misc.extension.extensionHook
 
-internal val mainActivityOnCreateHook = extensionHook {
-    strings("MainActivity_onCreate_begin")
+internal val messengerApplicationOnCreateHook = extensionHook {
```
Nuckyz, line 5: Why rename it?

patches/src/main/kotlin/app/revanced/patches/messenger/metaai/Fingerprints.kt
```diff
@@ -1,8 +1,11 @@
 package app.revanced.patches.messenger.metaai
 
 import com.android.tools.smali.dexlib2.Opcode
+import com.android.tools.smali.dexlib2.AccessFlags
 import app.revanced.patcher.fingerprint
 
+internal const val EXTENSION_METHOD_DESCRIPTOR = "Lapp/revanced/extension/messenger/metaai/RemoveMetaAIPatch;->overrideBooleanFlag(JZ)Z"
```
Nuckyz, line 7: Move this to the patch file

extensions/messenger/src/main/java/app/revanced/extension/messenger/metaai/RemoveMetaAIPatch.java
```diff
@@ -1,14 +1,21 @@
...
-        // This catches all flag IDs related to Meta AI.
-        // The IDs change slightly with every update,
-        // so to work around this, IDs from different versions were compared
-        // to find what they have in common, which turned out to be those first bits.
-        // TODO: Find the specific flags that we care about and patch the code they control instead.
-        if ((id & 0x7FFFFFC000000000L) == 0x810A8000000000L) {
+        if (Long.toString(id).startsWith("REPLACED_BY_PATCH")) {
+            synchronized (loggedIDs) {
+                if (loggedIDs.add(id)) {
+                    Logger.printInfo(() -> "Overriding " + id + " from " + value + " to false");
```
Nuckyz, line 16: Let it print everytime it overriddes, not just the first time. This way it's more clear when it's happening and allows to debug issues better

drobotk: I disagree, it's more useful to have it print them all out at once, instead of spam.

Issue #5182, opened 2025-06-15
feat: Pandora fix ads


Amby713:
### Feature description

Pandora has ads again, but unlimited skips still works. Can we get an updated patch brie the ads in Pandora?

### Motivation

Ads are the biggest problem. They were patched, but returned so we need new patch to fix it

### Acknowledgements

- [x] I have checked all open and closed feature requests and this is not a duplicate
- [x] I have chosen an appropriate title.
- [x] All requested information has been provided properly.

hoo-dles:
Does the patch fail? What version is it not working on?

Amby713:
Pandora version 2212.1 has ads now, but unlimited skip still works. The first time i patched it, ads went away. They made a change and ads came back. Patch needs to be updated

Amby713:
Patched version 2504.1.1 and updated. Played two songs and went to advertisement 

![Image](https://github.com/user-attachments/assets/61d8fff8-cedd-43f0-8082-10b33cb17738)

Amby713:
Closed by mistrust

hoo-dles:
@Amby713 there is a test build in the connected PR if you are able to patch via the CLI to confirm it works.

Amby713:
I've never used CLI, but have no problem. For now though, I can't set up for it until later this week. I'll update when I install it

PR #5185, opened 2025-06-16
fix(Pandora): Fix `Disable ads` patch

hoo-dles:
Addresses #5182

### DIFF
--- a/patches/src/main/kotlin/app/revanced/patches/pandora/ads/Fingerprints.kt
+++ b/patches/src/main/kotlin/app/revanced/patches/pandora/ads/Fingerprints.kt
@@ -1,0 +1,15 @@
+package app.revanced.patches.pandora.ads
+
+import app.revanced.patcher.fingerprint
+
+internal val isAdSupportedFingerprint = fingerprint {
+    custom { method, classDef ->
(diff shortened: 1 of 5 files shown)


patches/src/main/kotlin/app/revanced/patches/pandora/ads/Fingerprints.kt
```diff
@@ -0,0 +1,15 @@
+package app.revanced.patches.pandora.ads
+
+import app.revanced.patcher.fingerprint
+
+internal val isAdSupportedFingerprint = fingerprint {
+    custom { method, classDef ->
+        method.name == "getIsAdSupported" && classDef.endsWith("UserData;")
+    }
+}
```
oSumAtrIX, line 9:
```suggestion
internal val getIsAdSupportedFingerprint = fingerprint {
    custom { method, classDef ->
        method.name == "getIsAdSupported" && classDef.endsWith("UserData;")
    }
}
```

hoo-dles:
> The patch looks simple enough. If it breaks again though it can be constrained to a versionn

Patch should be backwards compatible too. The method I'm patching was in the previous version, so not sure what really changed. But seems to have stopped ads from propagating.

oSumAtrIX (approved):

PR #5242, opened 2025-06-23
fix(Spotify - Spoof Client): Ignore missing target files in hex patch

drobotk:


drobotk:
I was asked to to this by @oSumAtrIX on discord.

PR #5245, opened 2025-06-24
feat(YouTube - Hide video action buttons): Add `Hide Stop ads`

ILoveOpenSourceApplications:
Closes #5244.


ILoveOpenSourceApplications:
`Hide Save to playlist` should simply be called `Hide Save`. Why the extra info?

ILoveOpenSourceApplications:
I believe since this option is added for the first time, instead of quoting it, it should be inside a backwards quote.

PR #5318, opened 2025-07-01
fix(Spotify - Unlock Premium): Fix hiding ads context menus on latest versions

Nuckyz:


patches/src/main/kotlin/app/revanced/patches/spotify/misc/Fingerprints.kt
```diff
@@ -93,9 +118,9 @@ internal val abstractProtobufListEnsureIsMutableFingerprint = fingerprint {
     }
 }
 
-private fun structureGetSectionsFingerprint(className: String) = fingerprint {
+internal fun structureGetSectionsFingerprint(className: String) = fingerprint {
     custom { method, classDef ->
-        classDef.endsWith(className) && method.indexOfFirstInstruction {
+        classDef.type.endsWith(className) && method.indexOfFirstInstruction {
```
oSumAtrIX, line 123:
```suggestion
        classDef.endsWith(className) && method.indexOfFirstInstruction {
```

oSumAtrIX: Because its semantically duplicate. we want to check if the class is the one we search for. classDef.type is redundant.

oSumAtrIX: > must be used for equals or any string functions. 

Fair, i noticed this too, i suppose for consistency we can always make use of .type

PR #5392, opened 2025-07-09
fix(Sync for Reddit - Spoof client): Use www instead of ssl API to fix auth related issues

oSumAtrIX:


Beagon:
Works like a charm.

oSumAtrIX:
It already is

PR #5404, opened 2025-07-10
feat(Spotify): Remove support for old versions

Nuckyz:
It currently does not even patch correctly due to newly added extension hook fingerprints not matching. It's starting to become a burden to support for the little amount of users that might use it.


patches/src/main/kotlin/app/revanced/patches/spotify/lite/ondemand/OnDemandPatch.kt
```diff
@@ -1,21 +1,9 @@
 package app.revanced.patches.spotify.lite.ondemand
 
-import app.revanced.patcher.extensions.InstructionExtensions.addInstruction
 import app.revanced.patcher.patch.bytecodePatch
 
 @Deprecated("Patch no longer works and will be deleted soon")
 @Suppress("unused")
 val onDemandPatch = bytecodePatch(
```
oSumAtrIX, line 7: Is there a dependency on this patch somewhere

oSumAtrIX (approved):

PR #5483, opened 2025-07-16
fix(YouTube - Suggestion Provider): Correct suggestion provider string

sm455:


sm455:
This also fixes the issue of search suggestions not showing when not signed in.


After this patch.

markup_1000019172

oSumAtrIX (approved):
Awesome

ILoveOpenSourceApplications:
@sm455, does this also fix the issue of search history when not signed in or simply the suggestions?

sm455:
@ILoveOpenSourceApplications search history

Issue #5387, opened 2025-07-08
bug: Reddit Sync: Error Loading Page: 401


awesomeocelot:
### Bug description

Attempting to load any Reddit post in the Reddit Sync app fails with Error 401, this started occurring earlier today and based on this [Reddit post](https://www.reddit.com/r/revancedapp/comments/1luzl05/getting_an_error_loading_page_401_when_trying_to/) is affecting many Reddit Sync users.

I already have the latest Reddit Sync APK installed with the latest patches from the Revanced stable release, this error 401 appears to be due to a recent change on Reddit's end.

### Error logs

```shell

```

### Additional context

See https://www.reddit.com/r/revancedapp/comments/1luzl05/getting_an_error_loading_page_401_when_trying_to/

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] I have chosen an appropriate title.
- [x] All requested information has been provided properly.

kristianka:
Same issue, S21 FE 5G running Android 15. Regenerating an api token did not sadly work.

Seems like Reddit broke something server side. The question now is: is this intended to kill 3rd party apps, or a bug?

yashdes:
> Look at the other Redirect URI's from the [simplified guide](https://docs.google.com/document/u/0/d/1wHvqQwCYdJrQg4BKlGIVDLksPN0KpOnJWniT6PbZSrI/mobilebasic). Sync's is the only one that has a non-standard URL format `http://redditsync/auth`. Every other Redirect URI is in a valid format of `http://website.com/auth` or `protocol://example`. It could be Reddit just did a general validation on the URL's.

I think I would give some more credence to this if Relay is also broken

JoeyJoJoJr:
Patched RiF still works for browsing logged out, but gives "Server error: Try again later" when attempting to sign into your account. Possibly a clue to what they've changed?

Downy:
> Look at the other Redirect URI's from the [simplified guide](https://docs.google.com/document/u/0/d/1wHvqQwCYdJrQg4BKlGIVDLksPN0KpOnJWniT6PbZSrI/mobilebasic). Sync's is the only one that has a non-standard URL format `http://redditsync/auth`. Every other Redirect URI is in a valid format of `http://website.com/auth` or `protocol://example`. It could be Reddit just did a general validation on the URL's.

Continuum still works and uses a similar style URL (continuum://localhost)

codestation:
[...]

PR #5499, opened 2025-07-19
 fix(Relay for Reddit- Server error on login): Solved login issues while preserving offline browsing

ChielLaan:
Fixes #5497

## Problem
Relay for Reddit had errors during login after applying the spoof client patch. 
Because of some changed on Reddits side redirects stopped working if not exactly in the www.reddit.com structure , causing authentication failures similar to issue #5387 that affected Sync for Reddit.

On login you would get "server error" or "incorrect username/password" messages, while offline browsing continued to work normally. 
Blindly changing all `oauth.reddit.com` to `www.reddit.com` made API requests fail, so a more precise approach is needed.

## Solution
Applied a targeted OAuth domain redirect fix similar to the approach used for Sync for Reddit in #5387, but specifically tailored for Relay's authentication architecture:

- **`loginActivityClientIdFingerprint`** - OAuth URLs redirected from `oauth.reddit.com` to `www.reddit.com` (fixes login redirects)
- **`getLoggedInBearerTokenFingerprint`** - OAuth URLs redirected from `oauth.reddit.com` to `www.reddit.com` (handles authenticated token requests)
- **`getLoggedOutBearerTokenFingerprint`** - OAuth URLs **deliberately preserved** as `oauth.reddit.com` (maintains offline browsing functionality)

## Testing
Verified on Relay for Reddit 10.2.40 with custom client ID:

- **Login functionality**: Successfully authenticates without server errors or "incorrect username/password" messages
- **Offline browsing**: Continues to work when not logged in (preserves existing functionality)
- **Authenticated browsing**: Maintains full functionality when logged in

No additional dependent PRs.

### DIFF
--- a/patches/src/main/kotlin/app/revanced/patches/reddit/customclients/relayforreddit/api/SpoofClientPatch.kt
+++ b/patches/src/main/kotlin/app/revanced/patches/reddit/customclients/relayforreddit/api/SpoofClientPatch.kt
@@ -4,10 +4,12 @@ import app.revanced.patcher.extensions.InstructionExtensions.addInstructions
 import app.revanced.patcher.extensions.InstructionExtensions.getInstruction
 import app.revanced.patcher.extensions.InstructionExtensions.replaceInstruction
 import app.revanced.patches.reddit.customclients.spoofClientPatch
+import app.revanced.util.getReference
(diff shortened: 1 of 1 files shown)


oSumAtrIX:
Looking good, gonna do a deep review in a bit

Issue #4124, opened 2024-12-15
feat(Instagram): sanitize URLs and remove tracking parameters


bossanovaorca:
### Feature description

automatically remove Instagram URL trackers (e.g. https://instagram.com/p/[PHOTO_CODE]?igsh=[URL_TRACKER])

### Motivation

to prevent information logging and data collection when sharing a link

### Acknowledgements

- [x] I have checked all open and closed feature requests and this is not a duplicate
- [x] I have chosen an appropriate title.
- [x] All requested information has been provided properly.

KaeLL:
Up you go. This is sorely needed.

PR #3604, opened 2024-09-01
feat(Instagram): Added few patches

swakwork:
Hi team,

I have added the following patches to [Instagram](https://play.google.com/store/apps/details?id=com.instagram.android)

- Added `Enable developer menu` patch
- Added `Max media quality` patch
- Added `Selectable bio` patch
- Added `Sanitize sharing links` patch
- Added `Open links in external browser` patch

I have tested these with the latest alpha build(348.0.0.0.42) , at the time, and the patches work fine.

These patches go along with my integrations [PR #685](https://github.com/ReVanced/revanced-integrations/pull/685)

This is my first PR to this project, so do let me know if any code correction is needed.

Thanks,

PR #5986, opened 2025-09-24
feat(Instagram): Add `Sanitize sharing links` patch

brosssh:
Abstracted the extension method to sanitize links into the shared library, so it can be reused. Applied this new method for both Instagram and Spotify.

Closes https://github.com/ReVanced/revanced-patches/issues/4124.

### DIFF
--- a/extensions/instagram/src/main/java/app/revanced/extension/instagram/misc/privacy/SanitizeSharingLinksPatch.java
+++ b/extensions/instagram/src/main/java/app/revanced/extension/instagram/misc/privacy/SanitizeSharingLinksPatch.java
@@ -1,25 +1,22 @@
 package app.revanced.extension.instagram.misc.privacy;
 
-import android.net.Uri;
-import app.revanced.extension.shared.Logger;
+import app.revanced.extension.shared.privacy.SanitizeSharingLink;
 
 import java.util.List;
 
 @SuppressWarnings("unused")
 public final class SanitizeSharingLinksPatch {
-
     /**
      * Parameters that are considered undesirable and should be stripped away.
      */
     private static final List<String> SHARE_PARAMETERS_TO_REMOVE = List.of(
-            "si", // Share tracking parameter.
-            "utm_source" // Share source, such as "copy-link".
+            "igsh"
     );
 
     /**
      * Injection point.
      */
-    public static void sanitizeUrl(String url, String str2) {
-            Logger.printInfo(() -> "String1 " + url + " String2 " + str2);
+    public static String sanitizeSharingLink(String url) {
+        return SanitizeSharingLink.sanitizeUrl(url, SHARE_PARAMETERS_TO_REMOVE);
     }
 }
--- a/extensions/instagram/src/main/java/app/revanced/extension/instagram/misc/privacy/SanitizeSharingLinksPatch.java
+++ b/extensions/instagram/src/main/java/app/revanced/extension/instagram/misc/privacy/SanitizeSharingLinksPatch.java
@@ -1,0 +1,25 @@
+package app.revanced.extension.instagram.misc.privacy;
+
+import android.net.Uri;
+import app.revanced.extension.shared.Logger;
+
+import java.util.List;
+
+@SuppressWarnings("unused")
+public final class SanitizeSharingLinksPatch {
+
+    /**
+     * Parameters that are considered undesirable and should be stripped away.
+     */
+    private static final List<String> SHARE_PARAMETERS_TO_REMOVE = List.of(
+            "si", // Share tracking parameter.
+            "utm_source" // Share source, such as "copy-link".
+    );
+
+    /**
+     * Injection point.
+     */
+    public static void sanitizeUrl(String url, String str2) {
+            Logger.printInfo(() -> "String1 " + url + " String2 " + str2);
+    }
+}
--- a/extensions/shared/library/src/main/java/app/revanced/extension/shared/privacy/SanitizeSharingLink.java
+++ b/extensions/shared/library/src/main/java/app/revanced/extension/shared/privacy/SanitizeSharingLink.java
@@ -1,0 +1,30 @@
+package app.revanced.extension.shared.privacy;
+
+import android.net.Uri;
+import app.revanced.extension.shared.Logger;
+
+import java.util.List;
+
+public class SanitizeSharingLink {
+    public static String sanitizeUrl(String url, List<String> shareParameterToRemove) {
+        try {
+            Uri uri = Uri.parse(url);
+            Uri.Builder builder = uri.buildUpon().clearQuery();
+
+            for (String paramName : uri.getQueryParameterNames()) {
+                if (!shareParameterToRemove.contains(paramName)) {
+                    for (String value : uri.getQueryParameters(paramName)) {
+                        builder.appendQueryParameter(paramName, value);
+                    }
+                }
+            }
+
+            String sanitizedUrl = builder.build().toString();
+            Logger.printInfo(() -> "Sanitized url " + url + " to " + sanitizedUrl);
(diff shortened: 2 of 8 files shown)


brosssh:
Small nitpicking, should the name be "Sanitize sharing link**s**" (plural)? I sanitize only a link, but then it would be inconsistent with the other patch. Also, does it sanitize an url or a link? I updated Spotify to be consistent with the naming of the patch, so using `link`. Should the common method be called `sanitizeLink` instead of `sanitizeUrl`?

extensions/shared/library/src/main/java/app/revanced/extension/shared/privacy/SanitizeSharingLink.java
```diff
@@ -0,0 +1,30 @@
+package app.revanced.extension.shared.privacy;
+
+import android.net.Uri;
+import app.revanced.extension.shared.Logger;
+
+import java.util.List;
+
+public class SanitizeSharingLink {
```
oSumAtrIX, line 8: The class name should be suffixed with "Patch"

extensions/instagram/src/main/java/app/revanced/extension/instagram/misc/privacy/SanitizeSharingLinksPatch.java
```diff
@@ -0,0 +1,22 @@
...
+    private static final List<String> SHARE_PARAMETERS_TO_REMOVE = List.of(
+            "igsh"
+    );
+
+    /**
+     * Injection point.
+     */
+    public static String sanitizeSharingLink(String url) {
+        return SanitizeSharingLink.sanitizeUrl(url, SHARE_PARAMETERS_TO_REMOVE);
+    }
```
oSumAtrIX, line 21: Instead of doing this, make the shared class an abstract/interface and implement the parameters getter

oSumAtrIX: You can create a singleton instance field and access the implemented instance via a static method from a patch

extensions/shared/library/src/main/java/app/revanced/extension/shared/privacy/SanitizeSharingLink.java
brosssh: Well it's not really a patch tho at that place, it's just a utility method to strip out parameters from a string.

oSumAtrIX: Classes are object oriented, convert it to an interface/abstract

extensions/instagram/src/main/java/app/revanced/extension/instagram/misc/privacy/SanitizeSharingLinksPatch.java
oSumAtrIX: The class SanitizeSharingLink is not an object that matches the class content. It's clearly not a Link object.

extensions/shared/library/src/main/java/app/revanced/extension/shared/privacy/SanitizeSharingLinkPatch.java
```diff
@@ -0,0 +1,32 @@
+package app.revanced.extension.shared.privacy;
+
+import android.net.Uri;
+import app.revanced.extension.shared.Logger;
+
+import java.util.List;
+
+public abstract class SanitizeSharingLinkPatch {
+    protected abstract List<String> getParametersToRemove();
```
oSumAtrIX, line 9: Call this method and save it in a field, otherwise you call and instantiate a new list in O(n²) time below.

patches/src/main/kotlin/app/revanced/patches/reddit/misc/tracking/url/SanitizeUrlQueryPatch.kt
```diff
@@ -6,7 +6,7 @@ import app.revanced.patcher.patch.bytecodePatch
 @Suppress("unused")
 val sanitizeUrlQueryPatch = bytecodePatch(
```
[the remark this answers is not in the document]
brosssh: What is here and below?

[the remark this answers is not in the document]
oSumAtrIX: Check the Spoof client id patches for the various reddit clients as reference. Apply the same abstraction here.

Swakshan:
Hi I just wanted to check if this patch sanitizers urls of posts, reels, stories & live too ?.
If not can you check one of my previous PR and add them too https://github.com/ReVanced/revanced-patches/pull/3604/commits/6a54ffb17bc097ea73bc8c40faf60ee5cf681ad5#diff-d92b4bee79a91cc41c24ac6ca4da0d2945d8651c7ab0bc1d7a6784078aecab93 ( I got occupied and wasnt able to update the codebase that time)
Thanks :)

patches/src/main/kotlin/app/revanced/patches/reddit/misc/tracking/url/SanitizeUrlQueryPatch.kt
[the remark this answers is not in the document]
oSumAtrIX: Couple good reasons were mentioned in the Discord, including the possibility of converting them all into a universal patch. Regardless, the patches are shared in nature and thus should share code. The only diferences are the patch locations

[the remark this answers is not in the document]
brosssh: The only shared code is the name of the patch, and the the invoke static to an extension class (which has different name based on the package anyway). It adds additional complexity for a level of abstraction which is not needed. If you have a way to implement it without adding complexity, please propose code change. I can't find abstracting this useful in any way.

Also, how can it be a universal patch if you need to specify the method + register + index to inject? The universal patch suggested on Discord was about hooking the `ClipboardManager`, which this PR doesn't do..

extensions/instagram/src/main/java/app/revanced/extension/instagram/misc/privacy/SanitizeSharingLinksPatch.java
```diff
@@ -0,0 +1,25 @@
...
+@SuppressWarnings("unused")
+public final class SanitizeSharingLinksPatch extends SanitizeSharingLinkPatch {
+    private static final SanitizeSharingLinksPatch sanitizeSharingLinksPatch = new SanitizeSharingLinksPatch();
+
+    /**
+     * Parameters that are considered undesirable and should be stripped away.
+     */
+    @Override
+    protected List<String> getParametersToRemove() {
+        return List.of("igsh");
```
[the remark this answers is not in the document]
brosssh: I was told to do an abstract class with it https://github.com/ReVanced/revanced-patches/pull/5986#discussion_r2375533623.

brosssh (commented):
.

patches/src/main/kotlin/app/revanced/patches/reddit/misc/tracking/url/SanitizeUrlQueryPatch.kt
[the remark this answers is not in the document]
brosssh: @oSumAtrIX the PR is stuck for the reason in the comment thread. Please propose changes to apply abstraction without adding too much unneeded complexity or approve this. There is no reason to keep this in a limbo.

brosssh:
Shouldn't this kind of refactor be done in a different pr?

